PR #1, opened 2021-01-28
Error messages so sexy and descriptive, you won't belive it

FredTheDino:


src/error.rs
```diff
@@ -58,10 +61,19 @@ impl fmt::Display for ErrorKind {
...
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let message = match &self.message {
-            Some(s) => format!("\n{}", s),
+            Some(s) => format!("\n{} {}", ">>>".red(), s),
             None => String::from(""),
         };
-        write!(f, "{}:{} {}{}", self.file.display(), self.line, self.kind, message)
+
+        let line = if let Ok(file) = File::open(&self.file) {
+            io::BufReader::new(file).lines().enumerate()
```
sornas, line 69: enumerate starts at 0

PR #3, opened 2021-01-31
vectors

FredTheDino:
Adding in vectors/tuples, most basic operations work for them
and they are a nice way to store data. It's nice!


src/compiler.rs
```diff
@@ -331,6 +333,51 @@ impl Compiler {
...
 
+    fn grouping_or_tuple(&mut self, block: &mut Block) {
+        let block_length = block.ops.len();
+        let token_length = self.curr;
+        if self.try_tuple(block).is_err() {
+            block.ops.truncate(block_length);
+            self.curr = token_length;
+            self.grouping(block);
+        }
+    }
```
sornas, line 344: Could be a macro seeing as we do this in two places now. `maybe_else!(self.try_truple, self.grouping, block)`

src/compiler.rs
```diff
@@ -331,6 +333,51 @@ impl Compiler {
...
+                    if self.peek() == Token::RightParen {
+                        continue;
+                    }
+                    return Err(());
+                }
+            }
+        }
+        if num_args == 1 {
+            return Err(());
+        }
```
sornas, line 374: What about tuples with one element? Big no or..?

src/compiler.rs
```diff
@@ -331,6 +333,51 @@ impl Compiler {
...
+
+        let mut num_args = 0;
+        loop {
+            match self.peek() {
+                Token::RightParen | Token::EOF => {
+                    break;
+                }
+                Token::Newline => {
+                    self.eat();
+                }
```
sornas, line 357: Wild. Also, we need a language spec.

src/compiler.rs
```diff
@@ -331,6 +333,51 @@ impl Compiler {
...
+                    num_args += 1;
+                    if self.peek() == Token::Comma {
+                        self.eat();
+                        continue;
+                    }
+                    if self.peek() == Token::RightParen {
+                        continue;
+                    }
+                    return Err(());
+                }
```
sornas, line 369:
```suggestion
                token => {
                    self.expression(block);
                    num_args += 1;
                    match token {
                        Token::Comma => { self.eat(); }
                        Token::RightParen => {}
                        _ => { return Err(()); }
                    }
                }
```

src/compiler.rs
```diff
@@ -331,6 +333,51 @@ impl Compiler {
...
+            }
+        }
+        if num_args == 1 {
+            return Err(());
+        }
+
+        expect!(self, Token::RightParen, "Expected ')' after tuple");
```
sornas, line 376:
```suggestion
        expect!(self, Token::RightParen, "Expected ')' after tuple.");
```

src/compiler.rs
```diff
@@ -339,6 +386,15 @@ impl Compiler {
         expect!(self, Token::RightParen, "Expected ')' around expression.");
     }
 
+    fn index(&mut self, block: &mut Block) {
+        expect!(self, Token::LeftBracket, "Expected ']' around index.");
```
sornas, line 390:
```suggestion
        expect!(self, Token::LeftBracket, "Expected '[' around index.");
```

src/compiler.rs
```diff
@@ -339,6 +386,15 @@ impl Compiler {
...
+
+        self.expression(block);
+        block.add(Op::Index, self.line());
+
+        expect!(self, Token::RightBracket, "Expected '[' around index.");
```
sornas, line 395:
```suggestion
        expect!(self, Token::RightBracket, "Expected ']' around index.");
```

src/compiler.rs
```diff
@@ -472,6 +528,9 @@ impl Compiler {
         let mut return_type = Type::Void;
         let mut function_block = Block::new(&name, &self.current_file, self.line());
```
sornas, line 529: This line throws a compiler warning.

src/lib.rs
```diff
@@ -378,10 +387,114 @@ impl Debug for Value {
...
+    }
+
+    pub fn neg(value: &Value) -> Value {
+        match value {
+            Value::Float(a) => Value::Float(-a),
+            Value::Int(a) => Value::Int(-a),
+            Value::Tuple(a) => Value::Tuple(Rc::new(a.iter().map(neg).collect())),
+            _ => Value::Nil,
+        }
+    }
```
sornas, line 410: This should be its own PR.

src/lib.rs
```diff
@@ -378,10 +387,114 @@ impl Debug for Value {
...
+
+    pub fn not(value: &Value) -> Value {
+        match value {
+            Value::Bool(a) => Value::Bool(!a),
+            Value::Tuple(a) => Value::Tuple(Rc::new(a.iter().map(not).collect())),
```
sornas, line 415: I would love to hear a use case for this. I'm not hard-opposed, but I don't see the value. Or is it just to match that the other operators all work tuples? In any case, I think I prefer Python's `all` and the like (which are easy to expand with more variants) instead of extending the `!`-operator.

src/compiler.rs
FredTheDino: There is no syntax for it but it can be done. Tuples with one element might have uses.

src/lib.rs
FredTheDino: I see where you're comming from. But without the operators extracted to functions, the Tuple operations wouldn't work since they're recursive.

src/lib.rs
FredTheDino: I just added it for consistency. Since now an operator applied to a tuple, is the same as applying the operator to all it's elements. It's not super usefull since it requires the tuple to be bools, but I don't think this is an alternative to `all`. I see it as a supliment. Note that it returns a tuple.

src/compiler.rs
FredTheDino: Yes.

src/compiler.rs
FredTheDino: This code is not equivalent, but I like the idea.

src/compiler.rs
FredTheDino: I think this is a premature grouping. We might not always want to give `block` to all `try*` functions. There is a nice macro here, but that is a different PR.

src/lib.rs
sornas: Yes, but with this it's hard to see the Tuple additions since everything's additions.

src/lib.rs
sornas: Oh. That makes more sense actually.

src/compiler.rs
sornas: It isn't? I tried to make it equivalent.

sornas: Oh right I see it now.

src/vm.rs
```diff
@@ -119,12 +119,6 @@ impl VM {
         self.stack.push(value)
     }
 
-    fn pop_twice(&mut self) -> (Value, Value) {
-        let (a, b) = (self.stack.remove(self.stack.len() - 1),
-                      self.stack.remove(self.stack.len() - 1));
-        (b, a)  // this matches the order they were on the stack
-    }
-
```
sornas: Call this from two_op! instead of making two calls to pop and removing this.

sornas (approved):

Issue #36, opened 2021-02-01
Intern strings


sornas:

FredTheDino:
This will be irrelevant when we compile to lua.

PR #55, opened 2021-02-06
op copy

FredTheDino:
Changed the compiler to intern strings (in a dumb way) and
put constants in a list. This gave us a speedup of ~15%.
I ran some informal tests and got this:

before | after
-- | --
~1.0s | ~1.2s

I was running this program:
```
fib := fn a:int -> int {
    if a < 2 { ret a }
    ret fib(a - 1) + fib(a - 2)
}

fib(30)
```

I'd say that's a pretty nice speedup!

A stepping stone towards #36


sornas (approved):
Nice!

PR #56, opened 2021-02-06
parse branch macro

FredTheDino:
Sometimes we allow branches in the compiler, this is fine, but the way to do it is kinda cumbersome.

This PR adds a macro that generalises the branching, and makes it gives better error messages.

(I also fixed up the `run\_string!` macro a bit, it needed some love.)

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -946,7 +946,7 @@ impl Compiler {
         self.blobs.push(blob);
     }
 
-    fn try_blob_field(&mut self, block: &mut Block) -> Result<(), ()> {
+    fn blob_field(&mut self, block: &mut Block) {
         let name = match self.eat() {
             Token::Identifier(name) => name,
             _ => unreachable!(),
@@ -965,7 +965,7 @@ impl Compiler {
                             String::from(field)
                         } else {
                             error!(self, "Expected fieldname after '.'.");
-                            return Err(());
+                            return;
                         };
 
                         let op = match self.peek() {
@@ -973,7 +973,7 @@ impl Compiler {
                                 self.eat();
                                 self.expression(block);
                                 add_op(self, block, Op::Set(field));
-                                return Ok(());
+                                return;
                             }
 
                             Token::PlusEqual => Op::Add,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -734,7 +734,16 @@ mod tests {
...
             fn $fn() {
-                $crate::run_string($prog, true, Vec::new()).unwrap();
+                match $crate::run_string($prog, true, Vec::new()) {
+                    Ok(()) => {},
+                    Err(errs) => {
+                        for e in errs.iter() {
+                            println!("{}", e);
+                        }
+                        println!("  {} - FAILED\n", stringify!($fn));
+                        unreachable!();
```
sornas, line 744: minor: I think `panic!` is preferred

src/compiler.rs
```diff
@@ -39,6 +39,37 @@ macro_rules! expect {
...
+                    break;
+                }
+                $compiler.panic = false;
+                $compiler.curr = token_length;
+                let thrown_errors = $compiler.errors.len() - num_errors - 1;
+                stored_errors.extend($compiler.errors.split_off(thrown_errors));
+                $block.ops.truncate(block_length);
+            )*
+            break;
+        }
```
sornas, line 65: Why have a `while` if we always break when all `$call`s have been made?

src/compiler.rs
```diff
@@ -39,6 +39,37 @@ macro_rules! expect {
...
+        while !$compiler.panic {
+            $(
+                $call;
+                if !$compiler.panic {
+                    success = true;
+                    break;
+                }
+                $compiler.panic = false;
+                $compiler.curr = token_length;
+                let thrown_errors = $compiler.errors.len() - num_errors - 1;
```
sornas, line 60: Isn't `thrown_errors` re-declared in the same scope if multiple `$call` are passed? Unless this macro expansion is automatically scoped (which is kinda cool).

src/compiler.rs
FredTheDino: Because I needed a block to break out from. But didn't want to introduce a function in a macro. I just used my C-skills.

src/compiler.rs
FredTheDino: It is re-declared, but Rust allows that, shadowing the previous one. So the result is correct.

src/compiler.rs
sornas: Ah, I see.

src/compiler.rs
sornas: Indeed

sornas (approved):

sornas (approved):

Issue #48, opened 2021-02-05
The fancy call syntax


FredTheDino:
```
f!
```
Should call `f` with out any arguments.

```
f 1, 2
```
Calls 'f' with

PR #57, opened 2021-02-08
fancy call syntax

FredTheDino:
The original idea for the call syntax was:

```sylt
f := fn {}
f!
g := fn a:int {}
g 1
```

This created problems with `-1` being a valid expression and thus subtractions became function-calls.
We don't want that! So I settled on using this syntax:
```sylt
f := fn {}
f!
g := fn a:int {}
g! 1
```
Here the `!` is the call operator, very similar to the parentasis,
but you don't have to close it. :D
I hate closing things, like closing #48

Some duplicate commits from reworking some of the tests. Sorry about that.

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -579,29 +579,68 @@ impl Compiler {
     }
 
     fn call(&mut self, block: &mut Block) {
-        expect!(self, Token::LeftParen, "Expected '(' at start of function call.");
-
         let mut arity = 0;
-        loop {
-            match self.peek() {
-                Token::EOF => {
-                    error!(self, "Unexpected EOF in function call.");
-                    break;
-                }
-                Token::RightParen => {
-                    self.eat();
-                    break;
-                }
-                _ => {
-                    self.expression(block);
-                    arity += 1;
-                    if !matches!(self.peek(), Token::RightParen) {
-                        expect!(self, Token::Comma, "Expected ',' after argument.");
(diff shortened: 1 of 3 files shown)


src/compiler.rs
```diff
@@ -545,29 +545,68 @@ impl Compiler {
...
-                _ => {
-                    self.expression(block);
-                    arity += 1;
-                    if !matches!(self.peek(), Token::RightParen) {
-                        expect!(self, Token::Comma, "Expected ',' after argument.");
+            },
+
+            Token::Not => {
+                // I suspect this will be wierd with the boolean NOT
+                // operator...
```
sornas, line 578: Agree. Maybe rename it to `Token::Exclamation` or something.

src/compiler.rs
```diff
@@ -545,29 +545,68 @@ impl Compiler {
...
+                }
+                if !self.panic {
+                    println!("LINE {} -- ", self.line());
                 }
             }
-            if self.panic {
-                break;
+
+            _ => {
+                error!(self, "Not a valid function call, expected a '!' or a '('.");
```
sornas, line 609: Either "Invalid function call, expected" or "Not a function call. Expected"

sornas (commented):
I'm sort of impressed how small the actual change is. (Ignoring changes that are in #56 >:[ )

sornas (approved):

PR #60, opened 2021-02-10
break continue

FredTheDino:
Add in break and continue. They're not super tested, and there probably needs to be a some what more complex program written that
uses them. But I think I covered most of the bases.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -398,8 +398,8 @@ pub enum Op {
     /// {A} - JmpFalse(n) - {}
     JmpFalse(usize),
     /// Sets the instruction pointer
-    /// to the given value. And also pops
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -397,6 +397,14 @@ pub enum Op {
     ///
     /// {A} - JmpFalse(n) - {}
     JmpFalse(usize),
+    /// Sets the instruction pointer
+    /// to the given value. And also pops
+    /// the given number of values.
```
sornas, line 402:
```suggestion
    /// Sets the instruction pointer
    /// to the given value and pops
    /// the given amount of values.
```

sornas (approved):

PR #66, opened 2021-02-15
no-compilation order for constants

### LINKED ISSUE #52: Don't have compilation order in the year 2021
Nuff said.

FredTheDino:
Adds in functionality for compiling constant expressions in any order.
There are still some faulty parts, but the idea is there and it
can be looked at for improvements.

Closes #52

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -229,7 +229,7 @@ pub(crate) struct Compiler {
 
     blocks: Vec<Rc<RefCell<Block>>>,
     blob_id: usize,
-    unkowns: HashMap<String, (usize, usize)>,
+    unknown: HashMap<String, (usize, usize)>,
 
     functions: HashMap<String, (usize, RustFunction)>,
     constants: Vec<Value>,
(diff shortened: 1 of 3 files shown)


src/compiler.rs
```diff
@@ -621,10 +624,19 @@ impl Compiler {
...
-            .map(|(i, _)| i)
+    fn find_constant(&mut self, name: &str) -> usize {
+        let res = self.constants.iter().enumerate().find_map(|(i, x)| match x {
+            Value::Blob(b) if b.name == name => Some(i),
+            Value::Function(_, b) if b.borrow().name == name => Some(i),
+            _ => None,
+        });
+        if res.is_some() {
+            return res.unwrap();
+        }
```
sornas, line 635: `if let Some` ;)

FredTheDino:
I don't think we handle this kind of case correctly for all possible configurations:
```sylt
f()
q := 1
f :: fn {
    q += 1
}
```

sornas (approved):

Issue #63, opened 2021-02-10
Cleanup variable struct


FredTheDino:
Current the Variable struct is complex and noisy, it can probably be simplified a great deal.

FredTheDino:
I don't think there's much we can do without divine intervention. Closing this issue.

PR #67, opened 2021-02-16
unusued variables

### LINKED ISSUE #37: Catch unused variables
At least "variable X is never written to" and "variable X is never read"

FredTheDino:
(Added in a bunch of extra stuff, just because it made the PR better... 3432a32 and forrward is relevant.)

Makes it a compilation error if a variable is unused.

I'm unsure how I feel about this currently. It can get annoying,
but I think it's a good help and it needs to be available somehow.
The way to silence the error is to just, read the value on the line
after. It might be better to enforce a naming standard, like Rust?
Or maybe this is a good way to do it, since you don't have to rename
stuff?

Closes #37

Starts work on #63

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -229,7 +229,7 @@ pub(crate) struct Compiler {
 
     blocks: Vec<Rc<RefCell<Block>>>,
     blob_id: usize,
-    unkowns: HashMap<String, (usize, usize)>,
+    unknown: HashMap<String, (usize, usize)>,
 
     functions: HashMap<String, (usize, RustFunction)>,
     constants: Vec<Value>,
@@ -292,7 +292,7 @@ impl Compiler {
 
             blocks: Vec::new(),
             blob_id: 0,
-            unkowns: HashMap::new(),
+            unknown: HashMap::new(),
 
             functions: HashMap::new(),
 
@@ -635,7 +635,7 @@ impl Compiler {
         }
         let constant = self.add_constant(Value::Nil);
         let line = self.line();
-        let entry = self.unkowns.entry(name.to_string());
+        let entry = self.unknown.entry(name.to_string());
         entry.or_insert((constant, line)).0
     }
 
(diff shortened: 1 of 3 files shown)


FredTheDino:
killer feature: Allow the typechecker to run on code that is only faulty when it comes to unused variables. So a program can now be valid in different ways. Off to bed I go!

src/compiler.rs
```diff
@@ -241,10 +265,27 @@ macro_rules! push_frame {
...
+                        var.line,
+                        Token::Identifier(var.name.clone()
+                    ));
+                    $compiler.error_on_line(
+                        e,
+                        var.line,
+                        Some(format!("Usage of undefined value: '{}'.", var.name))
+                    );
+                }
+                $compiler.panic = false;
```
sornas, line 287: Maybe we'd like a proper warning!-macro instead of resetting the panic. Then it'd be easier to output "5 warnings found" or something.

src/compiler.rs
```diff
@@ -329,6 +394,26 @@ impl Compiler {
         &mut self.frames[last]
     }
 
+    /// Marks a variable as read, also marks upvalues.
```
sornas, line 397:
```suggestion
    /// Marks a variable as read. Also marks upvalues.
```

sornas (commented):
n> It might be better to enforce a naming standard, like Rust?
> Or maybe this is a good way to do it, since you don't have to rename
> stuff?

Do you mean something like the variable "_x" not being warned about? I'm unsure, actually. I wonder if we'd need to apply code inspection in more places than just here for it to feel like it fits. But it's a neat idea.

As for the renaming part I think it's not much of an issue. If it's unused you only need to change it in one place and it's not like there's a lot of names to choose from.

src/compiler.rs
sornas: Out of scope though.

FredTheDino:
> I think _some_ way of silencing the warning is needed before merging. Either renaming to `_x` or passing a hard-coded flag (as in, we don't need a full-blown `-Wno-unusued-variable`-system).

Not sure what the best solution is here. Passing flags to the compiler, will require a pretty large rewrite, and might be the better way. In some ways I dislike the Rust way, since it promotes a lot of edits to the code. But then again I guess people like it, because it makes it painfully obvious if a variable is unused.

Go enforces this as an error, making programs with unused variables faulty programs. We're not more extreme than Go here. And I think it's a good mentality to be extreme, and push for the features we want as hard as possible.

TL;DR I don't agree that we need a flag or way to silence the warning, other than making a "fake-usage" of the variable. We need to take a discussion here.

sornas (approved):

PR #71, opened 2021-02-17
better test fails

FredTheDino:
Fixed the test output, I wanted to add color but I had to do memory allocations and thought. Ceebs...

And now the faulty assert\_errs look like this

```
UNEXPECTED ERRORS
    GOT:  [ ErrorKind::SyntaxError(2, Int(2)) ]
    WANT: [ ErrorKind::InvalidProgram ]
```
Which is actually readable, by humans.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -840,11 +840,6 @@ mod tests {
         assert_errs!(run_string("a :: 2\nq :: fn { a = 2 }\n", true, Vec::new()), [ErrorKind::SyntaxError(_, _)]);
     }
 
-    #[test]
-    fn assign_to_constant_upvalue2() {
-        assert_errs!(run_string("a :: 2\nq :: fn { a = 2 }\n", true, Vec::new()), [ErrorKind::InvalidProgram]);
-    }
-
     macro_rules! test_multiple {
         ($mod:ident, $( $fn:ident : $prog:literal ),+ $( , )? ) => {
             mod $mod {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -797,8 +797,8 @@ mod tests {
                             for e in errs.iter() {
                                 println!("{}", e);
                             }
-                            println!("  {} - FAILED\n", stringify!($fn));
-                            panic!();
+                            eprintln!("  {} - FAILED\n", stringify!($fn));
+                            unreachable!();
                         }
                     }
                 });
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -730,17 +730,34 @@ mod tests {
...
     macro_rules! assert_errs {
         ($result:expr, [ $( $kind:pat ),* ]) => {
-            eprintln!("{} => {:?}", stringify!($result), $result);
-            assert!(matches!(
-                $result.unwrap_err().as_slice(),
+            let errs = if let Err(errs) = $result {
+                errs
+            } else {
+                eprintln!("    PROGRAM SUCCEEDED, WHEN IT SHOULD FAIL!");
+                unreachable!();
```
sornas, line 737: I don't think the tests should shout, even if they fail.
```suggestion
                eprintln!("    Program succeeded when it should've failed");
                unreachable!();
```

src/lib.rs
```diff
@@ -730,17 +730,34 @@ mod tests {
...
                 &[$($crate::error::Error {
                     kind: $kind,
                     file: _,
                     line: _,
                     message: _,
                 },
                 )*]
-            ))
+            ) {
+                eprintln!("UNEXPECTED ERRORS");
```
sornas, line 749:
```suggestion
                eprintln!("Unexpected errors");
```

src/lib.rs
```diff
@@ -730,17 +730,34 @@ mod tests {
...
+                eprint!("    GOT:  [");
```
sornas, line 750:
```suggestion
                eprint!("    Got:  [");
```

src/lib.rs
```diff
@@ -730,17 +730,34 @@ mod tests {
...
+                for err in errs {
+                    eprint!(" ErrorKind::{:?} ", err.kind);
+                }
+                eprint!("]\n    WANT: [");
```
sornas, line 754:
```suggestion
                eprint!("]\n    Want: [");
```

src/lib.rs
```diff
@@ -780,8 +797,8 @@ mod tests {
                             for e in errs.iter() {
                                 println!("{}", e);
                             }
-                            println!("  {} - FAILED\n", stringify!($fn));
-                            panic!();
+                            eprintln!("  {} - FAILED\n", stringify!($fn));
```
sornas, line 800:
```suggestion
                            eprintln!("  {} - failed\n", stringify!($fn));
```

sornas (approved):

PR #74, opened 2021-02-17
test suite

FredTheDino:
Add some more tests and update the bench-file.

Nothing much, but it should be enough to get information
about the performance of the interpeter.


FredTheDino:
This is part of the way to the target, but it's not as much as I want it to be. I would love to see some more tests.

progs/bench/fib_iter.sy
```diff
@@ -0,0 +1,14 @@
+// A fibonacci implementation that is a little
+// less awefull. But we run it 1000 times instead.
```
sornas, line 2:
```suggestion
// less awful. But we run it 1000 times instead.
```

progs/bench/fib.sy
```diff
@@ -0,0 +1,10 @@
+// The worst implementation of fibonacci calculations
```
sornas, line 1:
```suggestion
// The worst implementation of Fibonacci calculations
```

progs/bench/fib_iter.sy
```diff
@@ -0,0 +1,14 @@
+// A fibonacci implementation that is a little
```
sornas, line 1:
```suggestion
// A Fibonacci implementation that is a little
```

src/vm.rs
```diff
@@ -480,8 +480,9 @@ impl VM {
             self.frame().block.borrow().ops[self.frame().ip]);
     }
 
-    // Initalizes the VM for running. Run cannot be called before this.
-    pub(crate) fn init(&mut self, prog: &Prog) {
+    /// Initalizes the VM for running. Run cannot be called before this.
+    /// You do not have to call this.
+    pub fn init(&mut self, prog: &Prog) {
```
sornas, line 485: I feel like this should have a `#[doc(hidden)]` instead

sornas (approved):

PR #77, opened 2021-02-19
debug_print: only make one call to print

sornas:


FredTheDino (approved):
Thanks Gustav, very cool!

PR #92, opened 2021-02-24
Breaking into sections

FredTheDino:
The simple and dumb splitting of code into sections. So they can be compiled somewhat independently. There are still changes needed but it's a start.


src/lib.rs
```diff
@@ -842,7 +842,8 @@ mod tests {
             #[test]
             fn $fn() {
                 crate::tests::panic_after(std::time::Duration::from_millis(500), || {
-                    match $crate::run_string($prog, true, Vec::new()) {
+                    let prog = std::concat!("q :: fn {", $prog, "\n{}\n}\nq()");
+                    match $crate::run_string(&prog, true, Vec::new()) {
```
FredTheDino, line 846: This needs to be fixed for real. But it might be a good way to atleast use some of the already written tests.

sornas (approved):

Issue #100, opened 2021-03-07
Remove parse_branch


sornas:

FredTheDino:
What we actually want, is to remove parse_branch alltogether. It doesn't work like it should and there are other bugs from this.

PR #101, opened 2021-03-07
Single-element tuples

sornas:


progs/tests/tuples_fail_invalid_singleton.sy
```diff
@@ -0,0 +1,7 @@
+start :: fn {
+    a := (,)
+    a
+}
+
+//TODO(gu) See #100
+// errors: [ErrorKind::SyntaxError(_, _), ErrorKind::SyntaxError(_, _), ErrorKind::SyntaxError(_, _)]
```
sornas, line 7: #100

src/lib.rs
```diff
@@ -596,6 +601,7 @@ mod op {
             (Value::Int(a), Value::Int(b)) => Value::Int(a + b),
             (Value::String(a), Value::String(b)) => Value::String(Rc::from(format!("{}{}", a, b))),
             (Value::Tuple(a), Value::Tuple(b)) if a.len() == b.len() => tuple_bin_op(a, b, add),
+            (Value::Unknown, a) | (a, Value::Unknown) => a.clone(),
```
sornas, line 604: Swallowing a Value::Unknown this way is a bit sketchy. Ideally we'd only swallow when type checking but that's potentially _a lot_ of branching. `#[cfg(debug)]`?

FredTheDino: A nice fix here would be to pass in the tuple (a, a) for checking with this operation, so we don't have to assume the type supports the operation. And then only have the case `(Value::Unknown, Value::Unknown) => Value::Unknown`. Maybe?

progs/tests/tuples_fail_invalid_singleton.sy
FredTheDino: We should remove parse_branch.

FredTheDino (changes requested):
Looks good! But we can do slightly better type inference!

progs/tests/tuples_fail_invalid_singleton.sy
sornas: Well, yes. But I think it is out of scope here.

FredTheDino (approved):
LGTM

PR #111, opened 2021-03-11
fast and hacky performance counter

FredTheDino:
I made a fast and hacky performance counter. It did give a nice overview of the
runtime performance. I'm not sure we want this though, we at least want it extracted out
if we want to keep it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,7 @@ use std::fmt::Debug;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::hash::{Hash, Hasher};
+use std::io::Write;
 
 use owo_colors::OwoColorize;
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -49,6 +50,13 @@ fn run(args: Args, functions: Vec<(String, RustFunction)>) -> Result<(), Vec<Err
             if let Err(e) = vm.run() {
                 Err(vec![e])
             } else {
+                let mut log = std::fs::File::create("perf.csv").unwrap();
+                for (i, timer) in vm.times_per_op.iter().enumerate() {
+                    if let Some(name) = name_from_index(i) {
+                        let output = format!("'{}'| {}| {}|\n", name, timer.calls, timer.time);
```
sornas, line 56: You can pad with spaces here (`"{<22}|{>8}|{>12}"`) for prettiness. Also, `|` in a csv-file?

sornas (commented):
If we keep this I think we should make it optional. You don't want random .csv-files when running your game.

Also, a header for the csv would be nice.

sornas:
> we at least want it extracted out if we want to keep it.

I should read more carefully. Yes, fully agree.

src/lib.rs
FredTheDino: CSV allows any symbol to be used as a seprator. And we have commas in the Op::* names, since I was too lazy to remove them. So I picked a random character that I know didn't come up.

I like the padding idea though!

FredTheDino:
I wonder if we want to calculate all the data we might want from this? Or write a simple python script for generating the plots and displaying the information?

sornas:
matplotlib feels like a good fit.

PR #113, opened 2021-03-11
lists

FredTheDino:
Lists and type changes. A lot of small things happened, because they needed to happen.

I should maybe have split this up a bit more.


src/lib.rs
```diff
@@ -77,6 +78,9 @@ impl Default for Args {
 pub type RustFunction = fn(&[Value], bool) -> Result<Value, ErrorKind>;
 
 #[derive(Debug, Clone)]
+// TODO(ed): Our type comparison is wrong, we need something that is
+// asymetrical when comparing the types. Currently we just choose the
+// vauger of the types.
```
FredTheDino, line 83: #114 solves this comment.

src/lib.rs
```diff
@@ -168,14 +178,31 @@ impl From<&Value> for Type {
             Value::Tuple(v) => {
                 Type::Tuple(v.iter().map(|x| Type::from(x)).collect())
             }
+            Value::List(v) => {
+                let v: &RefCell<_> = v.borrow();
+                let v: &Vec<_> = &v.borrow();
+                let set: HashSet<_> = v.iter().map(|x| Type::from(x)).collect();
+                println!("{:?} ?> {:?}", v, set);
```
FredTheDino, line 185:
```suggestion
```

src/lib.rs
```diff
@@ -5,6 +5,7 @@ use std::fmt::Debug;
 use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::hash::{Hash, Hasher};
+use std::borrow::Borrow;
```
FredTheDino, line 8: I had to change all the code that used RefCell and Rc in this file, I don't really know why, and it annoys me. But this made the PR bigger than it should have been.

src/lib.rs
```diff
@@ -433,6 +476,11 @@ pub enum Op {
     ///
     /// {A, B, C} - Tuple(3) - {D(A, B, C)}
     Tuple(usize),
+    /// Creates a new [List] with the given size and place it on the top
+    /// of the stack.
+    ///
+    /// {A, B, C} - List(3) - {D(A, B, C)}
+    List(usize),
```
FredTheDino, line 483: I'm starting to wonder if this and Op::Tuple should be functions. Feels wierd to specify tuples here. But then again we probably get a speed increase.

src/main.rs
```diff
@@ -4,10 +4,11 @@ use sylt::{run_file, Args};
 
 fn main() -> Result<(), String> {
     let args = parse_args();
-    if args.file.is_none() {
-        return Err("No file to run".to_string());
-    }
-    let errs = match run_file(args, sylt_macro::link!(extern_test as test)) {
+    let errs = match run_file(args, sylt_macro::link!(
```
FredTheDino, line 7:
```suggestion
    if args.file.is_none() {
    let errs = match run_file(args, sylt_macro::link!(
        return Err("No file to run".to_string());
        sylt::dbg as dbg,
    }
    let errs = match run_file(args, sylt_macro::link!(
```

src/vm.rs
```diff
@@ -763,7 +781,7 @@ impl VM {
 
                         Value::ExternFunction(slot) => {
                             let extern_func = self.extern_functions[*slot];
-                            extern_func(&self.stack[new_base+1..], false)
+                            extern_func(&self.stack[new_base+1..], true)
```
FredTheDino, line 784: No idea why this was wrong honestly! We've never type checked external functions?

FredTheDino:
Also, the tests are failing. This is fixed by #114

src/lib.rs
sornas: I don't think we'll run into problems with "too many ops" any time soon.

src/main.rs
sornas: This suggestion is very hard to read.

src/compiler.rs
```diff
@@ -688,6 +689,44 @@ impl Compiler {
         parse_branch!(self, block, [self.tuple(block), self.grouping(block)]);
     }
 
+    fn list(&mut self, block: &mut Block) {
+        expect!(self, Token::LeftBracket, "Expected '[' at start of list.");
```
sornas, line 693:
```suggestion
        expect!(self, Token::LeftBracket, "Expected '[' at start of list");
```

src/compiler.rs
```diff
@@ -688,6 +689,44 @@ impl Compiler {
...
+        let mut num_args = 0;
+        loop {
+            match self.peek() {
+                Token::RightBracket | Token::EOF => {
+                    break false;
```
sornas, line 698:
```suggestion
                    break;
```

src/compiler.rs
```diff
@@ -688,6 +689,44 @@ impl Compiler {
...
+                    return;
+                }
+                _ => {
+                    self.expression(block);
+                    num_args += 1;
+                    match self.peek() {
+                        Token::Comma => {
+                            self.eat();
+                            if matches!(self.peek(), Token::RightParen) {
+                                break true;
```
sornas, line 714:
```suggestion
                                break;
```

src/compiler.rs
```diff
@@ -688,6 +689,44 @@ impl Compiler {
...
+        loop {
+            match self.peek() {
+                Token::RightBracket | Token::EOF => {
+                    break false;
+                }
+                Token::Newline => {
+                    self.eat();
+                }
+                Token::Comma => {
+                    error!(self, "Tuples must begin with an element or ')'.");
```
sornas, line 704:
```suggestion
                    error!(self, "Lists must begin with an element or ']'");
```

src/compiler.rs
```diff
@@ -688,6 +689,44 @@ impl Compiler {
...
+                    return;
+                }
+                _ => {
+                    self.expression(block);
+                    num_args += 1;
+                    match self.peek() {
+                        Token::Comma => {
+                            self.eat();
+                            if matches!(self.peek(), Token::RightParen) {
```
sornas, line 713:
```suggestion
                            if matches!(self.peek(), Token::RightBracket) {
```

src/compiler.rs
```diff
@@ -688,6 +689,44 @@ impl Compiler {
...
+                                break true;
+                            }
+                        },
+                        Token::RightBracket => {},
+                        _ => {
+                            error!(self, "Expected ',' or ']' after list element.");
```
sornas, line 719:
```suggestion
                            error!(self, "Expected ',' or ']' after list element");
```

src/compiler.rs
```diff
@@ -688,6 +689,44 @@ impl Compiler {
...
+                            return;
+                        },
+                    }
+                }
+            }
+        };
+        expect!(self, Token::RightBracket, "Expected ']' after list.");
```
sornas, line 726:
```suggestion
        expect!(self, Token::RightBracket, "Expected ']' after list");
```

src/compiler.rs
```diff
@@ -1481,6 +1520,16 @@ impl Compiler {
                 }
             }
 
+            Token::LeftBracket => {
+                self.eat();
+                let ty = self.parse_type();
+                expect!(self, Token::RightBracket, "Expected ']' after list type.");
```
sornas, line 1526:
```suggestion
                expect!(self, Token::RightBracket, "Expected ']' after list type");
```

src/lib.rs
sornas:
```suggestion
// vaguer of the types.
```

src/lib.rs
```diff
@@ -149,8 +158,10 @@ impl PartialEq for Type {
                 a.iter().zip(b.iter()).all(|(a, b)| a == b)
             }
             (Type::Union(a), b) | (b, Type::Union(a)) => {
+                // TODO(ed): This might be too lose, since b might be a union.
```
sornas:
```suggestion
                // TODO(ed): This might be too loose, since b might be a union.
```

src/lib.rs
```diff
@@ -220,13 +250,35 @@ impl From<Type> for Value {
     }
 }
 
+impl Type {
+    pub fn fits(&self, other: &Self) -> bool {
+        match (self, other) {
```
sornas, line 251: You could add a comment explaining what `fits` does since the name is a bit weird.

src/lib.rs
```diff
@@ -738,6 +807,18 @@ mod op {
...
+                let a = a.borrow();
+                let b = b.borrow();
+                if a.len() != b.len() {
+                    return Value::Bool(false);
+                }
+                a.iter().zip(b.iter()).find_map(
+                    |(a, b)| match eq(a, b) {
+                        Value::Bool(true) => None,
+                        a => Some(a),
+                    }).unwrap_or(Value::Bool(true))
```
sornas, line 816:
```suggestion
                Value::Bool(a.iter().zip(b.iter()).all(|(a, b)| matches!(eq(a, b), Value::Bool(true))))
```
This could be made a bit shorter if we made a `Value::try_inner` that's generic over the inner primitives.

Similar changes can be made for the other ops here.

sornas: Also, tuples can be written the same way.

src/lib.rs
```diff
@@ -1016,11 +1101,66 @@ mod tests {
...
+            let ls = Type::from(&ls[0]);
+            let v: Type = Type::from(&*v);
+            if ls == v {
+                Ok(Value::Nil)
+            } else {
+                Err(ErrorKind::TypeMismatch(ls, v))
+            }
+        }
+        ([Value::List(ls), v], false) => {
+            // NOTE(ed): Deliberattly no type checking.
```
sornas, line 1138:
```suggestion
            // NOTE(ed): Deliberately no type checking.
```

src/lib.rs
sornas: Something like this
```diff
diff --git a/src/lib.rs b/src/lib.rs
index 4a78ecc..349c315 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -268,6 +268,12 @@ impl Type {
     }
 }
 
+trait ToInner<T> {
+    fn to_inner(&self) -> Option<T>;
+
+    unsafe fn to_inner_unchecked(&self) -> T;
+}
+
 #[derive(Clone)]
 pub enum Value {
     Ty(Type),
@@ -289,6 +295,19 @@ pub enum Value {
     Nil,
 }
 
+impl ToInner<bool> for Value {
+    fn to_inner(&self) -> Option<bool> {
+        match self {
+            Value::Bool(b) => Some(*b),
+            _ => None,
+        }
+    }
+
+    unsafe fn to_inner_unchecked(&self) -> bool {
+        self.to_inner().unwrap_or_else(|| std::hint::unreachable_unchecked())
+    }
+}
+
 impl Debug for Value {
     fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
@@ -692,6 +711,7 @@ mod op {
     use super::{Type, Value};
     use std::rc::Rc;
     use std::collections::HashSet;
+    use crate::ToInner;
 
     fn tuple_bin_op(a: &Rc<Vec<Value>>, b: &Rc<Vec<Value>>, f: fn (&Value, &Value) -> Value) -> Value {
         Value::Tuple(Rc::new(a.iter().zip(b.iter()).map(|(a, b)| f(a, b)).collect()))
@@ -809,11 +829,7 @@ mod op {
                 if a.len() != b.len() {
                     return Value::Bool(false);
                 }
-                a.iter().zip(b.iter()).find_map(
-                    |(a, b)| match eq(a, b) {
-                        Value::Bool(true) => None,
-                        a => Some(a),
-                    }).unwrap_or(Value::Bool(true))
+                Value::Bool(a.iter().zip(b.iter()).all(|(a, b)| eq(a, b).to_inner().unwrap()))
             }
             _ => Value::Nil,
         }
```
I also made a `_unchecked` since this is to be used in the VM.

FredTheDino: This is not semantically equivalent. We can return Value::Nil from the comparison which means the comparison doesn't work because of the types. The type semantics have been changed here.

src/lib.rs
```diff
@@ -247,6 +247,8 @@ impl From<Type> for Value {
 }
 
 impl Type {
+    /// Checks if the other type is valid in a place where the self type is. It's an asymetrical
+    /// comparison for types usefull when checking assignment.
```
sornas, line 251:
```suggestion
    /// Checks if the other type is valid in a place where the self type is. It's an asymmetrical
    /// comparison for types useful when checking assignment.
```

src/lib.rs
```diff
@@ -793,11 +795,15 @@ mod op {
...
-                    }).unwrap_or(Value::Bool(true))
+                for (a, b) in a.iter().zip(b.iter()) {
+                    match eq(a, b) {
+                        Value::Bool(true) => {},
+                        Value::Bool(false) => { return Value::Bool(false); },
+                        Value::Nil => { return Value::Nil; },
+                        _ => unreachable!("Equality should only return bool or nil.")
+                    }
+                }
+                Value::Bool(true)
```
sornas, line 806: :+1:

src/lib.rs
```diff
@@ -1135,7 +1145,7 @@ pub fn push(values: &[Value], typecheck: bool) -> Result<Value, ErrorKind> {
             }
         }
         ([Value::List(ls), v], false) => {
-            // NOTE(ed): Deliberattly no type checking.
+            // NOTE(ed): Deliberatly no type checking.
```
sornas, line 1148:
```suggestion
            // NOTE(ed): Deliberately no type checking.
```

PR #114, opened 2021-03-11
type fits

FredTheDino:
Type fixes to make the type system stricter.

Is based on the list branch and ceebs rebasing.


src/lib.rs
```diff
@@ -250,6 +250,27 @@ impl From<Type> for Value {
     }
 }
 
+impl Type {
+    pub fn fits(&self, other: &Self) -> bool {
+        match (self, other) {
```
sornas, line 255: Unsure about the naming here. How can two types fit?

sornas: `assignable`?

FredTheDino: I agree that the naming is weird. I needed a name the communicates the asymmetry. I kinda want to use the "<=" operator, since that's how it works in my head, but it's not an exact match.

So it works like this:
```
Union(Int, Bool) fits Int
Int NOT fits Union(Int, Bool) 
```

I like the idea of assignable, but I read `assignable` as `mutable` or `modifyable`. I feel it brings the wrong connotations.

Maybe: `sub_type`, `contains`, `intersects`?

PR #123, opened 2021-03-12
better blob syntax

### LINKED ISSUE #121: Unify the blob syntax
This:
```
blob a {
    // Declare stuff...
}
```
Should be:

```
a :: blob {
    // Declare stuff...
}
```

FredTheDino:
Updates the blob-syntax to be unified with the rest of the language.
Just something small to keep me bussy while the array PR merges. ;)

Closes #121

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -1720,11 +1720,6 @@ impl Compiler {
                 add_op(self, block, Op::Yield);
             }
 
-            (Token::Identifier(_), Token::ColonColon, Token::Blob, ..) => {
-                self.blob_statement(block);
-            }
-
-
             (Token::Identifier(name), Token::ColonEqual, ..) => {
                 self.eat();
                 self.eat();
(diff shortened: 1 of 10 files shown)


src/compiler.rs
```diff
@@ -1518,13 +1518,14 @@ impl Compiler {
...
     fn blob_statement(&mut self, _block: &mut Block) {
-        expect!(self, Token::Blob, "Expected blob when declaring a blob");
         let name = if let Token::Identifier(name) = self.eat() {
             name
         } else {
             error!(self, "Expected identifier after 'blob'");
             return;
         };
+        expect!(self, Token::ColonColon, "Expected '::' when declaring a blob");
+        expect!(self, Token::Blob, "Expected blob when declaring a blob");
```
sornas, line 1528:
```suggestion
        expect!(self, Token::Blob, "Expected 'blob' when declaring a blob");
```

sornas (commented):
+1 for unifying syntax.

PR #125, opened 2021-03-13
smarter indexing

### LINKED ISSUE #116: Index should not return unknown
We can do better than Unknown when indexing! We also need tests to test this!

FredTheDino:
Fixes #116

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,22 @@
+use std::path::Path;
+
+fn find_test_paths(directory: &Path) {
+    println!("cargo:rerun-if-changed={}", directory.display());
+    for entry in std::fs::read_dir(directory).unwrap() {
+        let path = entry.unwrap().path();
+        let file_name = path.file_name().unwrap().to_str().unwrap();
+
+        if file_name.starts_with("_") {
+            continue;
+        }
+
+        if path.is_dir() {
+            find_test_paths(&path);
+        }
+    }
+}
+
+
 fn main() {
-    println!("rerun-if-changed=progs/")
+    find_test_paths(Path::new("progs/"));
 }
--- a/build.rs
+++ b/build.rs
@@ -1,3 +1,3 @@
 fn main() {
-    println!("cargo:rerun-if-changed=progs/")
+    println!("rerun-if-changed=progs/")
 }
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -1,3 +1,3 @@
 fn main() {
-    println!("cargo:rerun-if-changed=progs/")
+    println!("rerun-if-changed=progs/")
```
sornas, line 2: This _introduces_ the bug for me. With this change (i.e. without `cargo:`) the tests are recompiled if i `touch` README.md, but with `cargo:` intact it works like expected.

sornas: https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed If no `cargo:rerun-if-changed` is emitted then the project is always recompiled.

Have you tested that this change doesn't recompile if you only touch some unrelated file?

sornas (commented):
The code looks good though.

build.rs
FredTheDino: With `cargo:rerun-if-changed` changing and adding tests doesn't trigger a recompile. And I've wasted enough time writing tests that don't work.

I'd rather have false positives than false negatives.

FredTheDino: Also:
https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed
Says:
> If the path points to a directory, it will scan the entire directory for any modifications.

It might only go one level down. So:
 - A
     - B
         - D
     - C
 My problem is I change D and a recompile is not issued and it caches the old tests in the build file, giving me the same result. And that might be because it's only caching B and C. So we might have to change our build-script to look more like the test-finding thing.
 
 I understand your frustration, and I agree that it's annoying to always rebuild everything when all files are touched. But the problems of having the tests you write pass locally and fail on the CI is more annoying since it's invicible.

FredTheDino:
I tried to fix the buildscript. It now keeps track of all files in the progs/ tree. It's a little bit longer by not by much. And when I tested it I got no false negatives and no false positives. But it is more code.

As I see it, we either remove the build.rs file altogether. Or we use this more complex one.

sornas (approved):

PR #127, opened 2021-03-16
great speed

FredTheDino:


sornas (approved):
Look at this greedy diff algorithm. Unusable.

image

PR #129, opened 2021-03-16
Look in same directory when 'use'-ing. Also tests

sornas:


src/lib.rs
```diff
@@ -73,6 +73,13 @@ impl Default for Args {
     }
 }
 
+pub fn path_to_module(current_file: &Path, module: &str) -> PathBuf {
```
FredTheDino, line 76: Do you see this being used somewhere else? Why is it in lib?

progs/tests/import/main.sy
```diff
@@ -0,0 +1,6 @@
+use _other
```
FredTheDino, line 1: It would be nice with a test with circular dependencies, just to make sure.

src/sectionizer.rs
```diff
@@ -1,9 +1,11 @@
 use crate::error::{Error, ErrorKind};
 use crate::tokenizer::{PlacedToken, Token, file_to_tokens};
+use crate::path_to_module;
 
 use std::collections::HashSet;
 use std::path::{Path, PathBuf};
 
+#[derive(Debug)]
```
FredTheDino, line 8: :+1:

FredTheDino (approved):

src/lib.rs
sornas:
```
$ rg path_to_module src/**
src/sectionizer.rs
3:use crate::path_to_module;
54:                    let use_file = path_to_module(&path, file);

src/lib.rs
76:pub fn path_to_module(current_file: &Path, module: &str) -> PathBuf {

src/compiler.rs
6:use crate::{Blob, Block, Op, Prog, RustFunction, Type, Value, path_to_module};
1783:                    let path = path_to_module(file, &name);
```

FredTheDino (approved):
LGTM

PR #131, opened 2021-03-18
index assign

FredTheDino:
Now we can assign to indicies!

I also got feeling and added in initialization for the blobs, since the code was kinda broken and I had to change there anyways.

### DIFF
--- a/src/compiler.rs
+++ b/src/compiler.rs
@@ -1137,7 +1137,7 @@ impl Compiler {
                             Token::Equal => {
                                 self.eat();
                                 self.expression(block);
-                                add_op(self, block, Op::SetIndex);
+                                add_op(self, block, Op::AssignIndex);
                                 add_op(self, block, Op::Constant(nil));
                                 return;
                             }
@@ -1148,17 +1148,17 @@ impl Compiler {
                             Token::SlashEqual => Op::Div,
 
                             _ => {
-                                add_op(self, block, Op::Index);
+                                add_op(self, block, Op::GetIndex);
                                 continue;
                             }
                         };
 
                         add_op(self, block, Op::Copy(2));
-                        add_op(self, block, Op::Index);
+                        add_op(self, block, Op::GetIndex);
                         self.eat();
                         self.expression(block);
                         add_op(self, block, op);
-                        add_op(self, block, Op::SetIndex);
+                        add_op(self, block, Op::AssignIndex);
                         add_op(self, block, Op::Constant(nil));
                         return;
                     }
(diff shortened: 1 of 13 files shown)


src/compiler.rs
```diff
@@ -1135,6 +1124,44 @@ impl Compiler {
...
+
+                        let nil = self.add_constant(Value::Nil);
+                        let op = match self.peek() {
+                            Token::Equal => {
+                                self.eat();
+                                self.expression(block);
+                                add_op(self, block, Op::SetIndex);
+                                add_op(self, block, Op::Constant(nil));
+                                return;
+                            }
```
sornas, line 1143:
```suggestion
                        if self.peek() == Token::Equal {
                            self.eat();
                            self.expression(block);
                            add_op(self, block, Op::SetIndex);
                            add_op(self, block, Op::Constant(nil));
                            return;
                        }
                        
                        let op = match self.peek() {
```

src/compiler.rs
```diff
@@ -1598,14 +1625,46 @@ impl Compiler {
...
+                        self.expression(block);
+                        expect!(self, Token::RightBracket, "Expected ']' after indexing");
+
+                        let op = match self.peek() {
+                            Token::Equal => {
+                                self.eat();
+                                self.expression(block);
+                                add_op(self, block, Op::SetIndex);
+                                return;
+                            }
```
sornas, line 1647: As above

src/compiler.rs
```diff
@@ -1681,21 +1740,11 @@ impl Compiler {
                 self.assign(block);
             }
 
-            (Token::Identifier(_), Token::Dot, ..) => {
+            (Token::Identifier(_), Token::Dot, ..)
+                => {
```
sornas, line 1744:
```suggestion
            (Token::Identifier(_), Token::Dot, ..) => {
```

src/lib.rs
```diff
@@ -501,11 +501,16 @@ pub enum Op {
...
+    /// Indexes something indexable,
     /// and adds that element to the stack.
     ///
     /// {T, I} - Index - {T[I]}
     Index,
+    /// Sets the indexes of something indexable.
+    /// T[I] = V
+    ///
+    /// {T, I, V} - Index - {}
+    SetIndex,
```
sornas, line 513:
```suggestion
    /// Assigns to an index of something indexable.
    /// T[I] = V
    ///
    /// {T, I, V} - Index - {}
    AssignIndex,
```

src/vm.rs
```diff
@@ -756,12 +815,20 @@ impl VM {
                     let args: Vec<_> = args.iter().map(|x| x.into()).collect();
                     match callable {
                         Value::Blob(blob) => {
+                            if blob.fields.len() != num_args {
+                                return Err(ErrorKind::ArgumentCount(blob.fields.len(), num_args));
+                            }
```
sornas, line 820: A test for this would be nice.

PR #135, opened 2021-03-19
New macros

sornas:


sylt_macro/src/lib.rs
```diff
@@ -249,3 +249,69 @@ pub fn find_tests(tokens: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+    proc_macro::TokenStream::from(item)
+}
+
+#[proc_macro_derive(Next)]
+pub fn derive_next(item: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    assert!(!item.is_empty());
+    let parsed: syn::ItemEnum = parse_macro_input!(item);
+
+    let ident = parsed.ident.clone();
+    let match_arm: Vec<_> = parsed.variants.iter().zip(parsed.variants.iter().skip(1).chain(::std::iter::once(parsed.variants.iter().last().unwrap()))).map(|(var1, var2)| {
```
FredTheDino, line 301: :/
I honestly don't like it, I think it would be clearer to mutate the vector. Atleast break out `parsed.variants`, and if Rust doesn't let you, Rust is wrong.

FredTheDino (approved):
LGTM, except the iterator of doom

sylt_macro/src/lib.rs
sornas: How about this?

```rust
let mut iter = parsed.variants.iter();
let mut prev = iter.next().expect("Empty enum can't be Next");
let mut match_arms: Vec<_> = iter.map(|v| {
    let tokens = quote! {
        #ident::#prev => #ident::#v,
    };
    prev = v;
    tokens
}).collect();
match_arms.push(quote! {
    #ident::#prev => #ident::#prev,
});
```

sornas: Btw I just checked in godbolt and these two (but working with `u32`s) compiled to 65 and 66 instructions respectively.

<table>
<tr>
	<td> Amount of instructions
	<td> Iterator of doom
	<td> Sensible mutability
<tr>
	<td> Function
	<td> 56
	<td> 50
<tr>
	<td> Closure (map)
	<td> 9
	<td> 16
<tr> 
	<td> Total
	<td> 65
	<td> 66
</table>

sornas: I tried this as well (table updated above):
```rust
pub fn h(a: &[u32]) -> Vec<u32> {
    let mut prev = a.get(0).unwrap();
    let mut ret = Vec::new();
    for i in 1..a.len() {
        ret.push(prev + a[i]);
    }
    ret.push(prev + prev);
    ret
}
```
Now, I get that it isn't really the same when working with `u32`. It's just for fun :>

sornas: > Atleast break out parsed.variants, and if Rust doesn't let you, Rust is wrong.

Rust doesn't let me, since `.zip()` (and `.skip()`, and `.last()`, and `.map()`) takes ownership of the iterator (i.e. consumes it) and returns a new iterator. Even when borrowing the iterator (which is the case for e.g. `.next()`) it's a mutable borrow. The only function that takes a `&self` is `.size_hint()`.

FredTheDino (approved):

PR #138, opened 2021-03-20
sets and dicts

### LINKED ISSUE #23: Sets

### LINKED ISSUE #97: Dictionaries
It would be really nice with dictionaries.

FredTheDino:
A good start when it comes to sets and dicts. Somethings I'm thinking about:

~~I'm not sure about the syntax, since it's the same for sets and dicts, we don't know what
   an empty one is. And if we decide `{}` is an empty set, how do you make an empty dict?~~ Fixed

 - We might want to do a "contains" operator or instruction, for working with sets and dicts
 - We might want to change the return type from a dict to be `T?`, where `T` is the value type,
   since the look up might fail, and wanting to check if it's null makes sense but requires a dummy variable.
 - After looking at the code generated, we probably want a compile time optimizer that remove `Op::ThatPushes1Item, Op::Pop`, since those are wasted cycles and we can easily find this during compile time

Anyways, I'm going to go to sleep, I accidentally finished the feature. :3

Closes #23
Closes #97


src/lib.rs
```diff
@@ -247,7 +287,9 @@ impl Type {
         match (self, other) {
             (_, Type::Unknown) => true,
             (Type::List(a), Type::List(b)) => a.fits(b),
-            (Type::Union(a), Type::Union(b)) => a.iter().all(|x| b.contains(x)),
```
FredTheDino, line 250: I changed this to be stricter. Since all possible values we can get should be valid according to the type.

src/lib.rs
```diff
@@ -286,6 +330,8 @@ impl Debug for Value {
             Value::Bool(b) => write!(fmt, "(bool {})", b),
             Value::String(s) => write!(fmt, "(string \"{}\")", s),
             Value::List(v) => write!(fmt, "(array {:?})", v),
+            Value::Set(v) => write!(fmt, "(set {:?})", v),
+            Value::Dict(v) => write!(fmt, "(dict {:?})", v),
```
FredTheDino, line 332: These prints are a bit annoying. It would be great to just write the content, maybe a few borrows?

src/lib.rs
```diff
@@ -1120,13 +1178,15 @@ mod tests {
...
-);
+
+pub fn dbg(values: &[Value], _typecheck: bool) -> Result<Value, ErrorKind> {
+    println!("{}: {:?}, {:?}",
+        "DBG".purple(),
+        values.iter().map(Type::from).collect::<Vec<_>>(),
+        values
+    );
+    Ok(Value::Nil)
+}
```
FredTheDino: This was a godsent when debugging. It helped a lot, since it also prints what the typesystem is doing.

src/vm.rs
```diff
@@ -342,9 +371,9 @@ impl VM {
             }
 
             Op::AssignIndex => {
-                let value = self.stack.pop().unwrap();
-                let slot = self.stack.pop().unwrap();
-                let indexable = self.stack.pop().unwrap();
```
FredTheDino, line 347: We should always use the pop method.

sornas (approved):

src/lib.rs
```diff
@@ -469,16 +515,26 @@ pub enum Op {
...
     /// of the stack.
     ///
     /// {A, B, C} - List(3) - {D(A, B, C)}
     List(usize),
+    /// Creates a new [Value::Set] with the given elements and place it on the top
+    /// of the stack.
+    ///
+    /// {A, B, A} - Set(3) - {D(A, B)}
+    Set(usize),
+    /// Creates a new [Value::Dict] with the given elements and place it on the top
```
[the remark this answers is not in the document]
FredTheDino: Deliberate grouping.

progs/tests/set/returning.sy
```diff
@@ -0,0 +1,19 @@
...
+}
+
+start :: fn {
+    g :: q()
+    g[0] <=> true
+    false <=> g[1]
+
+    h :: h()
+    h[0] <=> true
+    true <=> h["a"]
```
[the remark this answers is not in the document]
sornas: #142

PR #140, opened 2021-03-21
Op::Force

### LINKED ISSUE #107: Asserting types to the compiler
Currently we allow this program:
```sylt
start :: fn {
    a: [int] = [1, "str", 3]
    ...
}
```
This is bad. We throw away type-information which might catch an error. And we might get a runtime error where we could have caught a compile time error. Thus we propose a change. The top program should not compile, because the types don't match in the declaration. But this new program will compile.

```sylt
start :: fn {
    a: [int]! = [1, "str", 3]
    ...
}
```
The exlamationpoint means "Assert type", telling the compiler that you promise it's the right type and it can ignore the type information it has. (Maybe this should enforce a runtime assertion to be run?)

FredTheDino:
Adds in the Op::Force, that can be used

```sylt
a :: (1, true)
b: !int = a[0]
```

Allows a lot of mischeif but it gives a way to compile all valid programs.

Closes #107

### DIFF
--- a/progs/tests/force/faulty.sy
+++ b/progs/tests/force/faulty.sy
@@ -1,7 +1,0 @@
-start :: fn {
-    a :: (1, "not", "an", "int")
-    b: int = a[0]
-    b
-}
-
-// errors: [ErrorKind::TypeMismatch(Type::Int, _)]
--- a/progs/tests/force/faulty.sy
+++ b/progs/tests/force/faulty.sy
@@ -1,0 +1,7 @@
+start :: fn {
+    a :: (1, "not", "an", "int")
+    b: int = a[0]
+    b
+}
+
+// errors: [ErrorKind::TypeMismatch(Type::Int, _)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -625,11 +625,11 @@ pub enum Op {
     /// {A} - AssignUpvalue(0) - {}
     AssignUpvalue(usize),
 
-    /// A helper instruction for the typechecker,
-    /// *makes sure* the top value on the stack
-    /// is of the given type, and is ment to signal
+    /// A helper instruction for the type checker.
+    /// *Makes sure* that the top value on the stack
+    /// is of the given type, and is meant to signal
     /// that the "variable" is added.
-    /// (The type is looked up in the constants vector)
+    /// (The type is looked up in the constants vector.)
     ///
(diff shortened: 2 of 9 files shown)


progs/tests/force/faulty.sy
```diff
@@ -0,0 +1,7 @@
+start :: fn {
+    a :: (1, "not", "an", "int")
+    b: int = a[0]
+    b
+}
+
+// errors: [ErrorKind::TypeMismatch(Type::Int, _)]
```
sornas, line 7: This is the same test as `force/faulty_tuple`.

src/lib.rs
```diff
@@ -626,13 +626,21 @@ pub enum Op {
     AssignUpvalue(usize),
 
     /// A helper instruction for the typechecker,
```
sornas, line 628:
```suggestion
    /// A helper instruction for the type checker.
```

src/lib.rs
```diff
@@ -626,13 +626,21 @@ pub enum Op {
...
-    /// makes sure the top value on the stack
+    /// *makes sure* the top value on the stack
     /// is of the given type, and is ment to signal
     /// that the "variable" is added.
     /// (The type is looked up in the constants vector)
```
sornas, line 632:
```suggestion
    /// *Makes sure* that the top value on the stack
    /// is of the given type, and is meant to signal
    /// that the "variable" is added.
    /// (The type is looked up in the constants vector.)
```

src/lib.rs
```diff
@@ -626,13 +626,21 @@ pub enum Op {
...
     ///
     /// Does not affect the stack.
     Define(usize),
+    /// A helper instruction for the typechecker,
+    /// *assumes* top value on the stack
+    /// is of the given type. Usefull for helping the
+    /// type system where it just can't do it.
+    /// (The type is looked up in the constants vector)
```
sornas, line 640:
```suggestion
    Define(usize),
    
    /// A helper instruction for the type checker.
    /// *Assumes* that the top value on the stack
    /// is of the given type. Useful for helping the
    /// type system where it just can't do it.
    /// (The type is looked up in the constants vector.)
```

sornas (approved):

src/lib.rs
FredTheDino: No, there's a grouping to this that is ruined by this formatting.

PR #144, opened 2021-03-21
contains

FredTheDino:
Added in the contains keyword for List, Set and Dict.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -559,6 +559,7 @@ pub enum Op {
     /// {O} - Get(F) - {O.F}
     Contains,
     /// Checks if the given value is inside the container.
+    /// Pushes a bool to the stack.
     ///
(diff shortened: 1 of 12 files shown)


src/lib.rs
```diff
@@ -557,6 +557,10 @@ pub enum Op {
     /// (name is looked up in the internal string-list)
     ///
     /// {O} - Get(F) - {O.F}
+    Contains,
+    /// Checks if the given value is inside the container.
```
sornas, line 561:
```suggestion
    /// Checks if the given value is inside the container.
    /// Pushes a bool to the stack.
```

sornas (approved):

src/lib.rs
FredTheDino: >

PR #146, opened 2021-03-22
iterators

FredTheDino:
Add fancypants iterators for great good!


FredTheDino:
needs tests with break and continue

src/compiler.rs
```diff
@@ -1492,6 +1492,41 @@ impl Compiler {
...
+        let inc = block.curr();
+        push_scope!(self, block, {
+            self.statement(block);
+        });
+        add_op(self, block, Op::Jmp(cond));
+
+        // patch_jmp!(Op::Jmp, cond_cont => block.curr());
+        block.patch(Op::Jmp(block.curr()), cond_cont);
+        self.scope(block);
+        add_op(self, block, Op::Jmp(inc));
```
sornas, line 1521: We should figure out if we want to do something like this macro or not. I still think that `block.patch()` is a bit hard to follow but maybe the alternative (something that looks more like `push_scope!`) is even harder to read.

src/lib.rs
```diff
@@ -233,7 +241,7 @@ impl From<&Value> for Type {
                 block.borrow().ty.clone()
             }
             Value::Unknown => Type::Unknown,
-            _ => Type::Void,
+            Value::Nil | Value::Ty(_) | Value::ExternFunction(_) => Type::Void,
```
sornas, line 244: :+1:

src/lib.rs
```diff
@@ -566,6 +577,17 @@ pub enum Op {
     /// {O} - Set(F) - {}
     AssignField(usize),
 
+    /// Turns the top element of the stack into an iterator.
+    ///
+    /// {A} - Iter - {Iter(A)}
+    Iter,
```
sornas, line 583:
```suggestion
    /// Turns the top element on the stack into an iterator.
    ///
    /// {A} - Iter - {Iter(A)}
    Iter,
```

src/lib.rs
```diff
@@ -566,6 +577,17 @@ pub enum Op {
...
+    /// Assumes the top element is an iterator. Steps the iterator
+    /// one step and pushes the value. If the iterator is consumed,
+    /// jump to the address and push [Value::Nil].
+    ///
+    /// {I} - JmpNext(line) - {I, A}
```
sornas, line 588:
```suggestion
    /// Steps the iterator on top of the stack one step and pushes
    /// the new value. If the iterator is consumed, jumps to the address
    /// and pushes [Value::Nil]. Errors if the top element isn't an iterator.
    ///
    /// {I} - JmpNext(line) - {I, A}
```

src/vm.rs
```diff
@@ -853,6 +914,50 @@ impl VM {
...
+            Op::JmpNext(_) => {
+                if let Some(Value::Iter(ty, _)) = self.stack.last() {
+                    let v = Value::from(ty);
+                    self.push(v);
+                } else {
+                    self.push(Value::Nil);
+                    error!(
+                        self,
+                        ErrorKind::InvalidProgram,
+                        "Can only 'next' iterators"
```
sornas, line 956: I'm guessing you mean to have different error messages so you can grep more easily and not have to care about if the error is when checking or running, but I think the solution is actually to report the errors differently. I've been thinking of something like this:

```rust
pub enum Error {
    CompileError {
        kind: ErrorKind,
        file: Option<PathBuf>,
        line: Option<usize>,
        message: Option<String>,
    },
    TypecheckError {
        kind: ErrorKind,
        file: PathBuf,
        line: usize,
        message: Option<String>,
    },
    RuntimeError {
        kind: ErrorKind,
        file: PathBuf,
        line: usize,
        message: Option<String>,
    },
    BincodeError,
}
```

We'd have to change some things in our test files, but that may be for the better since we'd have to specify *when* an error should be found. We don't want to miss when a type error used to be found during type checking but is suddenly only found in the runtime.

sornas: It's fine for this PR however, since we can't (easily) see a difference currently.

sornas (approved):

src/compiler.rs
sornas: I at least think we should switch the parameter ordering so it's read as "`patch` `cond_cont` to a `Op::Jmp(block.curr())`".

FredTheDino:
So... The (de)serialization is annoying because it needs to be implemented for the lambda-functions... Which is really annoying since it's not a type defined "here", so we need a dummy struct. And I cannot be asked to do that right now. Future me, pls fix.

PR #149, opened 2021-03-23
Clippy pass

sornas:
Most of these were style-changes. I skipped a few that I didn't agree with. We lost a few lines in the process.

The final two commits, however, might actually help performance.


FredTheDino (approved):
looks good. Maybe add to CI?

sornas:
Clippy has some peculiarities, so I'd rather run it manually once in a while.

PR #150, opened 2021-03-23
Compile to binary file

sornas:


src/lib.rs
```diff
@@ -23,15 +24,47 @@ pub trait Next {
...
+/// Compiles and serializes the given file.
+pub fn serialize(args: &Args, functions: Vec<(String, RustFunction)>) -> Result<Vec<u8>, Vec<Error>> {
+    let prog = compile(args, functions)?;
+    typecheck(&prog, args)?;
+    bincode::serialize(&prog).map_err(|_| vec![]) //TODO
+}
+
+/// Deserializes and links the given file.
+pub fn deserialize(bytes: Vec<u8>) -> Result<Prog, Vec<Error>> {
+    bincode::deserialize(&bytes).map_err(|_| vec![])
```
FredTheDino, line 63: 🤔 Are there really no errors here?

src/main.rs
```diff
@@ -1,40 +1,44 @@
...
     }
-    let errs = match run_file(
-        args,
-        sylt_macro::link!(sylt::dbg as dbg, sylt::push as push, sylt::len as len,),
-    ) {
-        Err(it) => it,
-        _ => return Ok(()),
+
+    let res = if args.is_binary {
+        let prog = sylt::deserialize(std::fs::read(args.file.clone().unwrap()).unwrap()).unwrap();
```
FredTheDino, line 18: the error here should bubble up.

src/main.rs
```diff
@@ -1,40 +1,44 @@
...
+                let mut dest = std::fs::File::create(compile_target).unwrap();
+                dest.write_all(&bytes).unwrap();
+                Ok(())
+            }
+            Err(e) => Err(e),
+        }
+    } else {
+        run_file(
+            &args,
+            sylt_macro::link!(sylt::dbg as dbg, sylt::push as push, sylt::len as len,),
```
FredTheDino, line 32: We should break out this function.

FredTheDino (approved):
LGTM, the errors need some love

PR #161, opened 2021-03-30
add a simple input handler

FredTheDino:
inputs! :D

### DIFF
--- a/lingon/src/input.rs
+++ b/lingon/src/input.rs
@@ -2,7 +2,8 @@ use luminance_sdl2::sdl2;
 use sdl2::event::{Event, WindowEvent};
 use std::collections::HashMap;
 
-pub use sdl2::controller::{Axis, Button};
+pub use sdl2::{GameControllerSubsystem, Sdl};
+pub use sdl2::controller::{Axis, Button, GameController};
 pub use sdl2::keyboard::Keycode;
 pub use sdl2::mouse::MouseButton;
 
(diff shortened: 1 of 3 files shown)


lingon/src/input.rs
```diff
@@ -0,0 +1,201 @@
...
+                    ..
+                } => {
+                    (Device::Quit, KeyState::Down(frame))
+                }
+                Event::KeyDown {
+                    keycode,
+                    ..
+                } => {
+                    if keycode.is_some() {
+                        (Device::Key(keycode.unwrap()), KeyState::Down(frame))
```
sornas, line 136:
```suggestion
                    if let Some(keycode) {
                        (Device::Key(keycode), KeyState::Down(frame))
```

lingon/src/input.rs
```diff
@@ -0,0 +1,201 @@
...
+                    } else {
+                        continue;
+                    }
+                }
+                Event::KeyUp {
+                    keycode,
+                    ..
+                } => {
+                    if keycode.is_some() {
+                        (Device::Key(keycode.unwrap()), KeyState::Up(frame))
```
sornas, line 146: dito

sornas (commented):
In this PR the windows ignores when I press shift+mod+c in my WM which previously closed the window. Pressing escape works.

FredTheDino:
> In this PR the windows ignores when I press shift+mod+c in my WM which previously closed the window. Pressing escape works.

I did dumdum, should be fixed! :)

sornas (approved):
:+1: works great

PR #166, opened 2021-04-07
crystalize compiler

FredTheDino:
Remove all the code that will be needed once we remove the runtime.

I also scrapped the typechecker, because I think I can write it better, without having to move between value space and type space.

This is an inital merge request in a series to compile sylt to rust. :)


sornas (approved):
image

Top ten saddest death scenes in anime.

PR #175, opened 2021-04-28
A simple start to the lingon bindings

FredTheDino:


src/lib.rs
```diff
@@ -635,6 +635,7 @@ pub enum Op {
     AssignField(usize),
 
     /// Turns the top element on the stack into an iterator.
+    /// Iter(Iter(A)) = Iter(A)
```
sornas, line 638:
```suggestion
    /// Turns the top element on the stack into an iterator.
    ///
    /// Iter(Iter(A)) = Iter(A)
```

src/lingon_sylt.rs
```diff
@@ -1,18 +1,113 @@
...
+fn new_game() -> Arc<Mutex<GG>> {
+    // TODO(ed): Maybe make these settable from the game itself.
+    Arc::new(Mutex::new(GG { game: Game::new("Lingon - Sylt", 500, 500) }))
+}
+
+macro_rules! game {
+    () => {
+        {
+            let mutex: &Mutex<GG> = Arc::borrow(&GAME);
+            &mut mutex.lock().unwrap().game
```
sornas, line 36:
```suggestion
            &mut GAME.lock().unwrap().game
```

src/lib_sylt.rs
```diff
@@ -53,4 +53,37 @@ sylt_macro::extern_function!(
...
+            let t: Type = Type::from(&*x);
+            let x = x.clone();
+            Ok(Value::Iter(t, Rc::new(RefCell::new(move || Some(x.clone())))))
+        }
+        values => Err(RuntimeError::ExternTypeMismatch(
+            "push".to_string(),
+            values.iter().map(Type::from).collect(),
+        )),
+    }
+}
```
sornas, line 86: From the name I assume this is supposed to be an infinite iterator, but when I test `example_game.sy` it runs for about one second before terminating.

sornas: I've been made aware of the `break`. Disregard.

sornas (approved):

src/lib_sylt.rs
sornas: I've been made aware of the `break` in the .sy. Disregard.

PR #176, opened 2021-04-29
lingon bindings input and rendering sprites and rendering sprites

FredTheDino:
All the nice prints a gal could possible need


Cargo.toml
```diff
@@ -18,7 +18,7 @@ sylt_macro = { path = "sylt_macro" }
 gumdrop = "0.8"
 
 criterion = { version = "0.3", optional = true }
-lingon = { git = "https://github.com/sornas/lingon.git", branch = "main", optional = true }
+lingon = { git = "https://github.com/sornas/lingon.git", branch = "fix-clone-for-real", optional = true }
```
sornas, line 21: Maybe we should pin this to a commit instead of the branch.

sornas (approved):

Cargo.toml
FredTheDino: I was hoping to remove it before the code was merged.

sornas: It's merged now *dab* but I still think we should pin to a specific commit. Otherwise someone might depend on this repository and it doesn't compile because lingon moved before sylt.

FredTheDino: We can fix that when we're not developing like it's nobodies business.

PR #199, opened 2021-05-01
Fix compiler crash with unknown fields on blobs

sornas:


FredTheDino:
We should always push Type::Unknown in the typechecker.

PR #200, opened 2021-05-01
Better macros

FredTheDino:
- Add MatchableValue
- Rewrite lib sylt
- Make the macro even more prettier
- Minor fixes

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -398,6 +398,35 @@ pub enum Value {
     Nil,
 }
 
+#[derive(Clone)]
+pub enum MatchableValue<'t> {
+    Empty,
+    One(&'t Value),
+    Two(&'t Value, &'t Value),
+    Three(&'t Value, &'t Value, &'t Value),
+    Four(&'t Value, &'t Value, &'t Value, &'t Value),
+    Five(&'t Value, &'t Value, &'t Value, &'t Value, &'t Value),
+}
+
+pub fn make_matchable<'t>(value: &'t Value) -> MatchableValue<'t> {
+    use crate::Value::*;
+    use MatchableValue::*;
+
+    match value {
+        Tuple(inner) => {
+            match (inner.get(0), inner.get(1), inner.get(2), inner.get(3), inner.get(4)) {
+                (Some(a), Some(b), Some(c), Some(d), Some(e), ..) => Five(a, b, c, d, e),
+                (Some(a), Some(b), Some(c), Some(d), ..) => Four(a, b, c, d),
+                (Some(a), Some(b), Some(c), ..) => Three(a, b, c),
+                (Some(a), Some(b), ..) => Two(a, b),
+                (Some(a), ..) => One(a),
+                _ => Empty,
+            }
+        },
+        x => One(x),
+    }
+}
+
 impl Debug for Value {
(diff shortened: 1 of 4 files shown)


FredTheDino:
It's the doc-test that fails... No idea why. One solution is changing the CI to `cargo test prog`.

It's frustrating tough -- the compiler complains about it doesn't understand what is what. But it's apparently not a problem when it's compiling. I almost suspect we've hit an edge case in the compiler.

sornas:
The compiler suggests two alternatives:

```
    = note: `sylt` could refer to a crate passed with `--extern`
    = help: use `::sylt` to refer to this crate unambiguously
note: `sylt` could also refer to the module imported here
   --> /home/gustav/dev/sylt/src/lib_sylt.rs:2:5
    |
2   | use crate as sylt;
    |     ^^^^^^^^^^^^^
    = help: use `self::sylt` to refer to this module unambiguously
```

and applying the second

```
diff --git a/sylt_macro/src/lib.rs b/sylt_macro/src/lib.rs
index 2649b25..b577016 100644
--- a/sylt_macro/src/lib.rs
+++ b/sylt_macro/src/lib.rs
@@ -77,7 +77,7 @@ pub fn extern_function(tokens: proc_macro::TokenStream) -> proc_macro::TokenStre
         let pat = block.pattern.clone();
         let ty = block.return_ty.clone();
         quote! {
-            #pat => { Ok(sylt::Value::from(#ty)) }
+            #pat => { Ok(self::sylt::Value::from(#ty)) }
         }
     }).collect();
 
@@ -95,27 +95,27 @@ pub fn extern_function(tokens: proc_macro::TokenStream) -> proc_macro::TokenStre
         pub fn #function (
             __values: &[sylt::Value],
             __typecheck: bool
-        ) -> ::std::result::Result<sylt::Value, sylt::error::RuntimeError>
+        ) -> ::std::result::Result<self::sylt::Value, self::sylt::error::RuntimeError>
         {
-            use sylt::Value::*;
-            use sylt::MatchableValue::*;
+            use self::sylt::Value::*;
+            use self::sylt::MatchableValue::*;
             if __typecheck {
                 let matching: Vec<_> = __values.iter().map(make_matchable).collect();
                 #[allow(unused_variables)]
                 match matching.as_slice() {
                     #(#typecheck_blocks),*
-                    _ => Err(sylt::error::RuntimeError::ExternTypeMismatch(
+                    _ => Err(self::sylt::error::RuntimeError::ExternTypeMismatch(
                         stringify!(#function).to_string(),
-                        __values.iter().map(|v| sylt::Type::from(v)).collect()
+                        __values.iter().map(|v| self::sylt::Type::from(v)).collect()
                     ))
                 }
             } else {
                 let matching: Vec<_> = __values.iter().map(make_matchable).collect();
                 match matching.as_slice() {
                     #(#eval_blocks),*
-                    _ => Err(sylt::error::RuntimeError::ExternTypeMismatch(
+                    _ => Err(self::sylt::error::RuntimeError::ExternTypeMismatch(
                         stringify!(#function).to_string(),
-                        __values.iter().map(|v| sylt::Type::from(v)).collect()
+                        __values.iter().map(|v| self::sylt::Type::from(v)).collect()
                     ))
                 }
             }
```
makes the tests pass.

PR #202, opened 2021-05-02
Add the documentation

FredTheDino:
- Add MatchableValue
- Rewrite lib sylt
- Make the macro even more prettier
- Minor fixes
----
Everything above this line is allready precent in "the other PR".
I couldn't be asked to integrate these changes one after another,
and I felt the needed to be done now. So I did them now. The diff might
be hard to look at though. :(

- Show where we need doc-strings
- Add some nice macro that writes to a file. :3
- Add a simple documentation generator
- I documented the lot!

### DIFF
--- a/src/lib_sylt.rs
+++ b/src/lib_sylt.rs
@@ -1,6 +1,7 @@
 use crate::*;
 use crate as sylt;
 
+#[sylt_macro::sylt_doc(dbg, "Writes the type and value of any thing you enter", [One(Value(val))] Type::Void)]
 #[sylt_macro::sylt_link(push, "sylt::lib_sylt")]
 pub fn dbg(values: &[Value], _typecheck: bool) -> Result<Value, RuntimeError> {
     println!(
@@ -12,6 +13,7 @@ pub fn dbg(values: &[Value], _typecheck: bool) -> Result<Value, RuntimeError> {
     Ok(Value::Nil)
 }
 
+#[sylt_macro::sylt_doc(push, "Appends an element to the end of a list", [One(List(ls)), One(Value(val))] Type::Void)]
 #[sylt_macro::sylt_link(push, "sylt::lib_sylt")]
 pub fn push(values: &[Value], typecheck: bool) -> Result<Value, RuntimeError> {
     match (values, typecheck) {
@@ -40,6 +42,7 @@ pub fn push(values: &[Value], typecheck: bool) -> Result<Value, RuntimeError> {
     }
 }
 
+#[sylt_macro::sylt_doc(prepend, "Adds an element to the start of a list", [One(List(ls)), One(Value(val))] Type::Void)]
 #[sylt_macro::sylt_link(prepend, "sylt::lib_sylt")]
 pub fn prepend(values: &[Value], typecheck: bool) -> Result<Value, RuntimeError> {
     match (values, typecheck) {
@@ -68,6 +71,7 @@ pub fn prepend(values: &[Value], typecheck: bool) -> Result<Value, RuntimeError>
     }
 }
 
+#[sylt_macro::sylt_doc(len, "Gives the length of tuples and lists", [One(Tuple(ls))] Type::Int, [One(List(ls))] Type::Int)]
 #[sylt_macro::sylt_link(len, "sylt::lib_sylt")]
 pub fn len(values: &[Value], _: bool) -> Result<Value, RuntimeError> {
     match values {
@@ -87,6 +91,7 @@ pub fn len(values: &[Value], _: bool) -> Result<Value, RuntimeError> {
 sylt_macro::extern_function!(
     "sylt::lib_sylt"
     sin
+    "The sinus function you know and love from trigenometry class"
     [One(Float(t))] -> Type::Float => {
         Ok(Float(t.sin()))
     },
@@ -95,6 +100,7 @@ sylt_macro::extern_function!(
 sylt_macro::extern_function!(
     "sylt::lib_sylt"
     cos
+    "The cosinus function you know and love from trigenometry class"
     [One(Float(t))] -> Type::Float => {
         Ok(Float(t.cos()))
     },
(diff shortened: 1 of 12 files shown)


src/lib_sylt.rs
```diff
@@ -1,6 +1,7 @@
 use crate::*;
 use crate as sylt;
 
+#[sylt_macro::sylt_doc(dbg, "Writes the type and value of any thing you enter", [One(Value(val))] Type::Void)]
 #[sylt_macro::sylt_link(push, "sylt::lib_sylt")]
 pub fn dbg(values: &[Value], _typecheck: bool) -> Result<Value, RuntimeError> {
```
sornas, line 6: The `push` here looks wrong. Maybe we should be able to catch it in the macro?

src/lib_sylt.rs
```diff
@@ -1,6 +1,7 @@
 use crate::*;
 use crate as sylt;
 
+#[sylt_macro::sylt_doc(dbg, "Writes the type and value of any thing you enter", [One(Value(val))] Type::Void)]
```
sornas, line 4:
```suggestion
#[sylt_macro::sylt_doc(dbg, "Writes the type and value of anything you enter", [One(Value(val))] Type::Void)]
```

src/lib_sylt.rs
```diff
@@ -68,41 +71,47 @@ pub fn prepend(values: &[Value], typecheck: bool) -> Result<Value, RuntimeError>
...
+        )),
+    }
+}
 
 sylt_macro::extern_function!(
     "sylt::lib_sylt"
     sin
-    [Value::Float(t)] -> Type::Float => {
-        Ok(Value::Float(t.sin()))
+    "The sinus function you know and love from trigenometry class"
```
sornas, line 94:
```suggestion
    "The sine function you know and love from trigonometry class"
```

src/lib_sylt.rs
```diff
@@ -68,41 +71,47 @@ pub fn prepend(values: &[Value], typecheck: bool) -> Result<Value, RuntimeError>
...
+        Ok(Float(t.sin()))
     },
 );
 
 sylt_macro::extern_function!(
     "sylt::lib_sylt"
     cos
-    [Value::Float(t)] -> Type::Float => {
-        Ok(Value::Float(t.cos()))
+    "The cosinus function you know and love from trigenometry class"
```
sornas, line 103:
```suggestion
    "The cosine function you know and love from trigonometry class"
```

src/lib_sylt.rs
```diff
@@ -68,41 +71,47 @@ pub fn prepend(values: &[Value], typecheck: bool) -> Result<Value, RuntimeError>
...
+        Ok(Float(t.cos()))
     },
 );
 
 sylt_macro::extern_function!(
     "sylt::lib_sylt"
     as_float
-    [Value::Int(t)] -> Type::Float => {
-        Ok(Value::Float(*t as f64))
+    "Converts the to a float"
```
sornas, line 112:
```suggestion
    "Converts the int to a float"
```

src/lib_sylt.rs
```diff
@@ -152,7 +161,7 @@ pub fn pop(values: &[Value], typecheck: bool) -> Result<Value, RuntimeError> {
     }
 }
 
-
+#[sylt_macro::sylt_doc(inf, "Returns an infinet iterator, spitting out the value you give it", [One(Value(val))] Type::Iter)]
```
sornas, line 164:
```suggestion
#[sylt_macro::sylt_doc(inf, "Returns an infinite iterator, spitting out the value you give it", [One(Value(val))] Type::Iter)]
```

src/lingon_sylt.rs
```diff
@@ -141,202 +72,280 @@ macro_rules! game {
 sylt_macro::extern_function!(
     "sylt::lingon_sylt"
     l_update
+    "Updates the engine, needs to be called once per frame"
```
sornas, line 75:
```suggestion
    "Updates the engine. Needs to be called once per frame"
```

src/lingon_sylt.rs
```diff
@@ -141,202 +72,280 @@ macro_rules! game {
...
         game!().update(0.0);
-        Ok(Value::Nil)
+        Ok(Nil)
     },
 );
 
 sylt_macro::extern_function!(
     "sylt::lingon_sylt"
     l_render
+    "Draws all render calls to the screen, needs to be called once per frame"
```
sornas, line 86:
```suggestion
    "Draws all render calls to the screen. Needs to be called once per frame"
```

src/lingon_sylt.rs
```diff
@@ -141,202 +72,280 @@ macro_rules! game {
...
+    [Two(Float(sx), Float(sy))] -> Type::Void => {
+        game!().renderer.camera.scale_by(*sx as f32, *sy as f32);
+        Ok(Nil)
     },
 );
 
 sylt_macro::extern_function!(
     "sylt::lingon_sylt"
     l_gfx_particle_new
+    "Creates a new particle system. Note specially the return type, don't edit the return value."
```
sornas, line 275:
```suggestion
    "Creates a new particle system. Note specially the return type. Don't edit the return value."
```

src/lingon_sylt.rs
```diff
@@ -397,102 +418,114 @@ particle_prop! { l_gfx_particle_end_alpha, end_alpha }
...
         use lingon::input::{Device, Button};
         let button = if let Some(button) = Button::from_string(button) {
             button
         } else {
             return error!("l_bind_button", "'{}' is an invalid button", button);
         };
 
-        game!().input.bind(Device::Button(*controller as u32, button), String::clone(name));
-        Ok(Value::Nil)
+        game!().input.bind(Device::Button(*controller as u32, button), std::string::String::clone(name));
```
sornas, line 502: Consider `name.clone()` instead, both here and above/below.

sornas (approved):
What about the failing CI?

src/lib_sylt.rs
FredTheDino: That was what the "no duplicates" were supposed to catch...

src/lingon_sylt.rs
FredTheDino: I couldn't. And I got frustrated....

PR #215, opened 2021-05-17
Export gumdrop

sornas:


FredTheDino:
No idea if it works - but it is a nice idea! :3

FredTheDino (approved):

sornas:
I ran into this with #216. You'll see that I had to `use gumdrop::Options` in the new binary as well. I tried this over there and it worked, but I wanted to split it.

PR #216, opened 2021-05-18
Time runtime and do some statistical analysis

sornas:
It's a bit rough around the edges but the logic is all there.

Motivation: Timing the entire binary also counts compilation and file IO, which isn't really what we want. With this we test only the runtime of the program and do our own statisticial analysis (kinda) using scipy.

```sh
~/dev/sylt $ cargo run --release --bin time -- -r 100 progs/bench/fib_iter.sy > stats1
    Finished release [optimized] target(s) in 0.03s
     Running `target/release/time -r 100 progs/bench/fib_iter.sy`
Compiling
Running once
Starting runs
~/dev/sylt $ cargo run --release --bin time -- -r 100 progs/bench/fib_iter.sy > stats2
    Finished release [optimized] target(s) in 0.03s
     Running `target/release/time -r 100 progs/bench/fib_iter.sy`
Compiling
Running once
Starting runs
~/dev/sylt $ wc -l stats1 stats2
 100 stats1
 100 stats2
 200 total
~/dev/sylt $ python time_analysis.py stats1 stats2
Mean
29749.880
30175.730

Standard deviation
1181.843
1629.033

P values (small values -> reject equal mean hypothesis):
Equal variance   Unequal variance
0.036            0.036
~/dev/sylt $ 
```

We could look into https://github.com/PyO3/pyo3 and do the analysis in Rust as well.

### DIFF
--- a/src/bin/time.rs
+++ b/src/bin/time.rs
@@ -1,14 +1,11 @@
 use gumdrop::Options;
-use std::{io::{Write, stdout}, path::PathBuf};
+use std::path::PathBuf;
 use std::time::Instant;
 
 #[derive(Default, Debug, Options)]
 pub struct Args {
     #[options(free, required, help = "The file to run")]
     pub run_file: PathBuf,
-
-    #[options(free, help = "The file to write the runtimes to (omit or '-' for stdout)")]
-    pub stat_file: Option<PathBuf>,
 
     #[options(short = "r", long = "runs", help = "If set, how many times the program should be run at most")]
     pub max_runs: Option<u32>,
@@ -83,18 +80,8 @@ fn main() -> std::io::Result<()> {
         runtimes.push(runtime.as_micros());
     }
 
-    eprintln!("Saving to file");
-    match args.stat_file {
-        Some(file) => write_stats(&mut std::fs::File::create(file).unwrap(), &runtimes)?,
-        None => write_stats(&mut stdout(), &runtimes)?,
-    }
-
-    Ok(())
-}
-
-fn write_stats<W: Write>(to: &mut W, stats: &[u128]) -> std::io::Result<()> {
-    for stat in stats {
-        writeln!(to, "{}", stat)?;
+    for runtime in runtimes {
+        println!("{}", runtime);
     }
     Ok(())
 }
--- a/src/bin/time.rs
+++ b/src/bin/time.rs
@@ -8,7 +8,7 @@ pub struct Args {
     pub run_file: PathBuf,
 
     #[options(short = "r", long = "runs", help = "If set, how many times the program should be run at most")]
-    pub max_runs: Option<u32>,
+    pub max_runs: Option<usize>,
 
     #[options(short = "t", long = "time", help = "If set, stop running when this amount of seconds have passed")]
     pub max_time: Option<u64>,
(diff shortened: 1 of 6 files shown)


src/bin/time.rs
```diff
@@ -0,0 +1,87 @@
...
+                eprintln!("{}", err);
+            }
+            return Ok(());
+        }
+    }
+    let mut runtimes = Vec::new();
+
+    eprintln!("Starting runs");
+    let outer_start = Instant::now();
+    let mut runs = 0;
```
FredTheDino, line 63: Isn't this implicit from the length of the vector?

src/bin/time.rs
```diff
@@ -0,0 +1,87 @@
...
+        runs += 1;
+
+        let start = Instant::now();
+        let _ = sylt::run(&prog, &sylt_args).unwrap();
+        let runtime = Instant::now() - start;
+        runtimes.push(runtime.as_micros());
+    }
+
+    for runtime in runtimes {
+        println!("{}", runtime);
```
FredTheDino, line 84: Guessing you do the math yourself? Maybe clearer begin and ends?

Great that you used stderr and stdout.

time_analysis.py
```diff
@@ -0,0 +1,34 @@
+import sys
+
+try:
+    from scipy import stats as scipy_stats
+except ModuleNotFoundError:
+    print("Need scipy for statistical analysis", file=sys.stderr)
+    sys.exit(1)
```
FredTheDino, line 7: I see you're a professional.

time_analysis.py
```diff
@@ -0,0 +1,34 @@
+import sys
```
FredTheDino, line 1: Maybe allow piping from stdin?

time_analysis.py
```diff
@@ -0,0 +1,34 @@
...
+print()
+print("Standard deviation")
+for stat in stats:
+    print(f"{stat.variance**(1/2):.3f}")
+print()
+# Normal p-values are p<0.05 or even p<0.01 for statistical significance.
+# Perhaps try a baseline where you compare the programs with themselves.
+print("P values (small values -> reject equal mean hypothesis):")
+print("Equal variance   Unequal variance")
+print(f"{p_eq_var:.3f}            {p_uneq_var:.3f}")
```
FredTheDino, line 34: Haven't run it by it looks good!

There are some things that might be useful that isn't recorded. Maybe store hope many instructions the program is and the like, so we can compare compiler optimizations aswell?

We might be a little bit more markup in the data file, maybe a really dumb json writer that just blurts out the content? But that doesn't have to go on straight away.

How does this work for games? Can we make it useful there somehow? Counting frames maybe? Or until the yield keyword?

FredTheDino (commented):
Were cool! I wanna try it before i approve by i like it!

time_analysis.py
sornas: Yeah this currently only makes sense for non-games. But I think game performance should focus more on metrics like "average frame time" and the likes while this focuses solely on "how long does this program take to terminate" and relies on /many/ runs to compare runtimes.

src/bin/time.rs
sornas: Unsure what you mean here. `eprintln!("Runtimes follow:");` won't really work since `stdout` and `stderr` aren't synchronized. (AFAIK?)

sornas: But yes, the math is done in the python.

PR #220, opened 2021-05-30
syntax trees building

FredTheDino:
21-06-02
I've matched the old syntax as tightly as possible - there are however some small TODOs that need 
addressing, but I would like to test them out before removing them.

The parser doesn't find any syntax errors in test-suite we have made, except for the for-loops.
Other than that the parser seams to work pretty great! I can't think of anything more to test
or anything more to remove, so I think this is done for now.

For some highlevel documentation, there are 4 functions to worry about:
 - statement, does all statement parsing
 - expression, has a lot of internal functions for passing around expressions, but that is only available internally
 - parse_type, parses types with some fancy recursion (a lot simpler than the old one)
 - function, parsing functions, it's a pretty big deal with functions, but it is a lot simpler than it used to be

### DIFF
--- a/src/syntree.rs
+++ b/src/syntree.rs
@@ -168,7 +168,7 @@ pub enum ExpressionKind {
     // Simple
     Float(f64),
     Int(i64),
-    String(String),
+    Str(String),
     Bool(bool),
     Nil,
 }
@@ -312,7 +312,7 @@ fn expression<'t>(ctx: Context<'t>) -> ParseResult<'t, Expression> {
                 T::Int(i) => Int(i),
                 T::Bool(b) => Bool(b),
                 T::Nil => Nil,
-                T::String(s) => String(s),
+                T::String(s) => Str(s),
                 t => {
                     raise_syntax_error!(ctx, "Cannot parse value, '{:?}' is not a valid value", t);
                 }
@@ -486,8 +486,8 @@ pub fn construct(tokens: &Tokens) -> Result<Module, Vec<Error>> {
 mod test {
     use crate::tokenizer::string_to_tokens;
     use super::*;
-    use StatementKind::*;
     use ExpressionKind::*;
+    use AssignableKind::*;
 
     macro_rules! test_expression {
         ($name:ident: $str:expr => $ans:pat) => {
@@ -509,4 +509,19 @@ mod test {
     }
 
     test_expression!(simple_expr: "0" => Expression { kind: Int(0), .. });
+    test_expression!(simple_add: "0 + 1.0" => Expression { kind: Add(_, _), ..  });
+    test_expression!(simple_mul: "\"abc\" * \"abc\"" => Expression { kind: Mul(_, _), ..  });
+    test_expression!(simple_ident: "a" => Expression {
+        kind: Get(Assignable { kind: Read(_), .. }),
+        ..
+    });
+    test_expression!(simple_access: "a.b" => Expression {
+        kind: Get(Assignable { kind: Access(_, _), .. }), ..
+    });
+    test_expression!(simple_index_ident: "a[a]" => Expression {
+        kind: Get(Assignable { kind: Index(_, _), .. }), ..
+    });
+    test_expression!(simple_index_expr: "a[1 + 2 + 3]" => Expression {
+        kind: Get(Assignable { kind: Index(_, _), .. }), ..
+    });
 }
--- a/src/syntree.rs
+++ b/src/syntree.rs
@@ -1020,13 +1020,26 @@ mod test {
 
         test!(parse_type, type_dict_one: "{int : int}" => Dict(_, _));
         test!(parse_type, type_dict_complex: "{int | float? : int | int | int?}" => Dict(_, _));
+
+        // TODO(ed): Require block or allow all statements?
+        test!(expression, simple: "fn -> {}" => _);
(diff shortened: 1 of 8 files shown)


src/syntree.rs
```diff
@@ -1,7 +1,7 @@
-use std::path::PathBuf;
+use std::path::{PathBuf, Path};
 use std::collections::HashMap;
 use crate::error::Error;
-use crate::tokenizer::TokenStream;
+use crate::tokenizer::{TokenStream, Token};
 use super::Type as runtimeType;
```
sornas, line 5: This is a type, so

```suggestion
use super::Type as RuntimeType;
```

src/syntree.rs
```diff
@@ -189,6 +189,115 @@ pub struct Type {
...
+    };
+}
+
+macro_rules! raise_syntax_error {
+    ($ctx:expr, $( $msg:expr ),* ) => {
+        return Err(($ctx.skip(1), vec![syntax_error!($ctx, $( $msg ),*)]))
+    };
+}
+
+fn expression<'t>(ctx: Context<'t>) -> Result<(Context<'t>, Expression), (Context<'t>, Vec<Error>)> {
```
sornas, line 269: What about `-> (Context<'t>, Result<Expression, Vec<Error>>)` instead? Same for `outer_statement`.

sornas (commented):
I tried running this but got more errors than tokens in the file, any idea what's up with that?

FredTheDino:
> I tried running this but got more errors than there are tokens in the file, any idea what's up with that?

Basically nothing is implemented. It compiles numbers right now. Nothing more fancy than that. This is about the only valid program:

```
1
1.0
```

src/syntree.rs
FredTheDino:
```rust
285 |     let (ctx, value) = expression(ctx)?;
    |                        ^^^^^^^^^^^^^^^^ the `?` operator cannot be applied to type `(syntree::Context<'_>, std::result::Result<Expression, Vec<error::Error>>)`
``` 
I think it's worth keeping it like it is for egonomics with the `? ` operator. But it was a good idea!

src/syntree.rs
```diff
@@ -189,6 +189,115 @@ pub struct Type {
...
-pub fn construct(tokens: TokenStream) -> Result<Prog, Vec<Error>> {
-    Err(Vec::new())
+type Tokens = [(Token, usize)];
+
+#[derive(Debug, Copy, Clone)]
+struct Context<'a> {
+    pub tokens: &'a Tokens,
+    pub curr: usize,
+    pub file: &'a Path,
+}
```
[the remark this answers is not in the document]
FredTheDino: Unsure what you mean. I'm guessing you wonder why I made it a borrow instead of moving in the `PathBuf` into the struct? I wanted to make `Context` copy if I could, and since all of it is pointers that don't own any of the memory, a copy is perfectly legal. This is nice since we clone it alot when parsing, so we always can revert back.

src/syntree.rs
```diff
@@ -1,188 +1,1263 @@
...
+                    let (ctx, ty) = parse_type(ctx)?;
+                    let kind = match (ctx.token(), banger) {
+                        (T::Colon, T::Bang) => VarKind::ForceConst,
+                        (T::Equal, T::Bang) => VarKind::ForceMutable,
+                        (T::Colon, _) => VarKind::Const,
+                        (T::Equal, _) => VarKind::Mutable,
+                        (t, _) => { raise_syntax_error!(ctx, "Expected ':' or '=' for definition, but got '{:?}'", t); }
+                    };
+                    (ctx.skip(1), kind, ty)
+                }
```
sornas, line 552:
```suggestion
                T::Colon => {
                    let ctx = ctx.skip(1);
                    let forced = matches!(ctx.token(), T::Bang);
                    let ctx = skip_if!(ctx, T::Bang);
                    let (ctx, ty) = parse_type(ctx)?;
                    let kind = match (ctx.token(), forced) {
                        (T::Colon, true) => VarKind::ForceConst,
                        (T::Equal, true) => VarKind::ForceMutable,
                        (T::Colon, false) => VarKind::Const,
                        (T::Equal, false) => VarKind::Mutable,
                        (t, _) => { raise_syntax_error!(ctx, "Expected ':' or '=' for definition, but got '{:?}'", t); }
                    };
                    (ctx.skip(1), kind, ty)
                }
```

src/syntree.rs
```diff
@@ -1,188 +1,1263 @@
...
+        }
+    };
+
+    // TODO(ed): Not sure this is right.
+    // let ctx = expect!(ctx, T::Newline, "Expected newline after statement");
+    let ctx = skip_if!(ctx, T::Newline);
+    Ok((ctx, Statement { span, kind }))
+}
+
+fn maybe_call<'t>(ctx: Context<'t>, calle: Assignable) -> ParseResult<'t, Assignable> {
```
sornas, line 595:
```suggestion
fn maybe_call<'t>(ctx: Context<'t>, callee: Assignable) -> ParseResult<'t, Assignable> {
```

sornas (commented):
About halfway through. Will continue later.

sornas (approved):
The review actually worked, even though it's +1338 ;)

Tests left:

- [ ] unary
- [ ] bools
- [x] comparisons
- [x] some x-assign operators
- [x] `-` and `/` as infix operators

PR #229, opened 2021-06-17
Syntax trees final touchups

sornas:
- [ ] Final renames
- [x] Unignore ignored tests
- [x] Fix tests (global declaration order added to list instead, all other tests run)
- [x] Outer statements
- [x] `rg TODO`
- [ ] `f!` -> `f'`
- [x] Test the game jam game (https://github.com/sornas/ultimate-fishbee/tree/sylt-syntax-trees)
  - [x] No errors occured (55 left)
  - [x] `l := [\n1\n]` reproduced
  - [x] `if !a && b {` reproduced
  - [x] `a! b[c], d\n}` can't reproduce (but still fixed)
  - [x] Stack ordering in the typechecker seems to be wrong sometimes (typecheck errors like "Extern function x doesn't accept arguments with type(s) ...) (this was due to multiplie upvalues being captured)
- [ ] Lots of typechecking propagation when a void-type is produced. Ignore it?
- [x] Generate syntax errors for conflict markers
- [x] Catch globals declaration order mismatch
- [ ] "Invalid expression" points to the wrong token.
- [x] "No valid expression" sometimes points to the wrong line number (this was due to reading the wrong file)
- [x] Update path when reading other files


src/syntree.rs
```diff
@@ -1213,7 +1211,7 @@ mod expression {
         Ok((ctx, Expression { span, kind }))
     }
 
-    /// Parse something that begins at the start of a expression.
+    /// Parse something that begins at the start of an expression.
```
FredTheDino, line 1215: :eyes:

FredTheDino (approved):

PR #251, opened 2021-06-29
Less error propogation

### LINKED ISSUE #231: Lots of typechecking propagation when a void-type is produced. Ignore it?
If a type error occures, lots of errors propogate from it. This makes it hard to read the output and causes headaches.

One solution could be to add a new type: Faulty. The Faulty-type will never trigger an error. So errors don't propogate.

FredTheDino:
Closes #231

Adds invalid types that signals to the typechecker what errors can be ignored.


sylt-common/src/ty.rs
```diff
@@ -84,6 +86,10 @@ impl Hash for Type {
             Type::ExternFunction(_) => {
                 16
             }
+
+            Type::Invalid => {
+                19
+            }
```
sornas, line 92: I don't like that these numbers are un-ordered.

sornas (approved):

sylt-common/src/ty.rs
FredTheDino: I have a todo to make the enum "countable" - that would remove the entire match block and would make me happy. I might do that now actually.

FredTheDino: See #255

FredTheDino: See #255

PR #253, opened 2021-06-30
Allow newlines in more places

### LINKED ISSUE #193: Allow newlines in inner expressions
These should be valid:
```
a := (
	1 + 
    2 +
    3
)

b := f( 1,
2, 3
,4)
```

FredTheDino:
Closes #193

Handle newlines in wierd places - though this change isn't trubble free.

There's a special case that I haven't thought of, consider:
```
A {
    a: 2,
    b: 3,
}
```
This could be either:
```
A        - Identifier
{ .. }   - A dict initlization
```
or
```
A { .. } - A blob initlization
```

Dosn't know which currently - but it makes me think I might have done goofed.
Are sets and dicts legal after any identifier? And how will this affect the `block expressions`?

### DIFF
--- a/progs/tests/blob/ambiguous_if.sy
+++ b/progs/tests/blob/ambiguous_if.sy
@@ -1,11 +1,11 @@
 A :: blob {
-    a: int
-    b: [int]
+    a: int,
+    b: [int],
 }
 
 start :: fn {
     a := A {
-        b: [1, 2]
-        a: 0
+        b: [1, 2],
+        a: 0,
     }
 }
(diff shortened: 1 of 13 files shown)


sornas:
> Are sets and dicts legal after any identifier?

They can always be parsed as two statement expressions after each other.

> And how will this affect the block expressions?

Yeah, that's a good question. Semicolons might be the answer.

sornas (approved):
I don't want to manually check all calls to `{push,pop}_skip_newlines` but the logic looks good.

Issue #169, opened 2021-04-19
The ops modules should always return results


FredTheDino:
Currently we return `Type::Void` when an operation isn't allowed. It's a bit janky, but can give us some bad type-errors. We could guess what type we should return and return that, which would give better type-errors. Now we kinda get a series of faults from 1 error.

FredTheDino:
Isn't this solved by #254 ?

PR #254, opened 2021-06-30
Number all type variants

FredTheDino:
Adds a macro so enums can be converted into numbers - very pain free.

### DIFF
--- a/sylt-common/src/ty.rs
+++ b/sylt-common/src/ty.rs
@@ -3,7 +3,11 @@ use std::hash::{Hash, Hasher};
 
 use crate::{Value, Blob};
 
-#[derive(Debug, Clone)]
+pub trait Numbered {
+    fn to_number(&self) -> usize;
+}
+
+#[derive(Debug, Clone, sylt_macro::Numbered)]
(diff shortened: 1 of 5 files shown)


sylt-common/src/ty.rs
```diff
@@ -29,67 +33,43 @@ pub enum Type {
...
+            Type::Ty
+            | Type::Void
+            | Type::Unknown
+            | Type::Int
+            | Type::Float
+            | Type::Bool
+            | Type::ExternFunction(..)
+            | Type::String => {}
         }
         .hash(h);
```
sornas, line 74: Should this line still be here?

sornas (approved):
This looks good but I'm a bit confused about the PRs now. #255 looks like a duplicate  of this and I think the linked issue here (#169) should link to something else.

sylt-common/src/ty.rs
FredTheDino: Yes

FredTheDino, line 74:
```suggestion
```

PR #257, opened 2021-07-01
Implicit returns

sornas:


sylt-parser/src/expression.rs
```diff
@@ -146,9 +146,36 @@ fn function<'t>(ctx: Context<'t>) -> ParseResult<'t, Expression> {
...
     // Parse the function statement.
-    let (ctx, statement) = block_statement(ctx)?;
+    let (ctx, mut statement) = block_statement(ctx)?;
+
+    // If the return type isn't void, check for and apply implicit returns.
+
+    let statements = if let StatementKind::Block { statements } = &mut statement.kind {
+        statements
+    } else {
+        unreachable!("Function blocks should only be blocks");
```
FredTheDino, line 157: Do we want to allow expressions? :o

progs/tests/implicit_return/simple.sy
```diff
@@ -0,0 +1,17 @@
...
+
+h :: fn a:int, b:int -> int {
+    a + b
+}
+
+start :: fn {
+    f() <=> 1
+    g() <=> 2
+    h(1, 2) <=> 3
+}
```
FredTheDino, line 17: Maybe add more tests?

FredTheDino (approved):
LGTM

sylt-parser/src/expression.rs
sornas: Maybe? Not here at least

PR #273, opened 2021-07-04
if expressions 2

FredTheDino:
Adds some nice little if-expressions!

### DIFF
--- a/progs/tests/expression/if_expressions.sy
+++ b/progs/tests/expression/if_expressions.sy
@@ -1,0 +1,24 @@
+grouping :: fn {
+    a := 1 + 2 + 3 if true else -1 * 10 if false else 2 + 2
+    a <=> 4
+    b := 1 + 2 + 3 if true else -1 * 10 if true else 2 + 2
+    b <=> 6
+    c := 1 + 2 + 3 if false else -1 * 10 if true else 2 + 2
+    c <=> -10
+    d := 1 + 2 + 3 if false else -1 * 10 if false else 2 + 2
+    d <=> 4
+}
+
+simple :: fn {
+    (1 if true else 3) <=> 1
+}
+
+expr :: fn {
+    (1 + (1 if true else 3)) <=> 2
+}
+
+start :: fn {
+    grouping'
+    simple'
+    expr'
+}
--- a/progs/tests/expression/if_expressions.sy
+++ b/progs/tests/expression/if_expressions.sy
@@ -1,6 +1,6 @@
 grouping :: fn {
     a := 1 + 2 + 3 if true else -1 * 10 if false else 2 + 2
-    a <=> 4
+    a <=> 6
     b := 1 + 2 + 3 if true else -1 * 10 if true else 2 + 2
     b <=> 6
     c := 1 + 2 + 3 if false else -1 * 10 if true else 2 + 2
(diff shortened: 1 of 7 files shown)


progs/tests/expression/if_expressions.sy
```diff
@@ -0,0 +1,24 @@
+grouping :: fn {
+    a := 1 + 2 + 3 if true else -1 * 10 if false else 2 + 2
+    a <=> 4
+    b := 1 + 2 + 3 if true else -1 * 10 if true else 2 + 2
+    b <=> 6
+    c := 1 + 2 + 3 if false else -1 * 10 if true else 2 + 2
+    c <=> -10
+    d := 1 + 2 + 3 if false else -1 * 10 if false else 2 + 2
+    d <=> 4
+}
```
sornas, line 10: I actually don't agree with this grouping. I read these like this (parenthesis added for clarity):

```
    a := 1 + 2 + (3 if true else (-1 * (10 if false else (2 + 2))))
    a <=> 6
    b := 1 + 2 + (3 if true else (-1 * (10 if true else (2 + 2))))
    b <=> 6
    c := 1 + 2 + (3 if false else (-1 * (10 if true else (2 + 2))))
    c <=> -7
    d := 1 + 2 + (3 if false else (-1 * (10 if false else (2 + 2))))
    d <=> -1
```
I think this is clearer since here, if the condition is true, you know that the rest of the expression isn't necessary. (Unless you have parenthesis.)

sylt-common/src/op.rs
```diff
@@ -235,13 +235,21 @@ pub enum Op {
     /// Does not affect the stack.
     Define(usize),
     /// A helper instruction for the typechecker,
-    /// *assumes* top value on the stack
+    /// *Assumes* top value on the stack
```
sornas, line 238: Either a small `a` in `assumes` and the comma above, or a capital `A` in `Assumes` and a period above.

sylt-common/src/op.rs
```diff
@@ -235,13 +235,21 @@ pub enum Op {
...
     /// is of the given type. Usefull for helping the
     /// type system where it just can't do it.
     /// (The type is looked up in the constants vector)
     ///
     /// Does not affect the stack.
     Force(usize),
+    /// A helper instruction for the typechecker,
+    /// *Combines* the two top value on the stack
```
sornas, line 246: Same here for `Combines`, either `c` and comma or `C` and period.

sylt-common/src/op.rs
```diff
@@ -235,13 +235,21 @@ pub enum Op {
...
+    /// into a union type.
+    ///
+    /// Skipped in the runtime.
+    /// In typechecker
+    /// {A, B} - Union - {A | B}
```
sornas, line 251:
```suggestion
    /// In the typechecker:
    /// {A, B} - Union - {A | B}
```

progs/tests/expression/if_expressions.sy
FredTheDino: I don't know how nice this is to read - when they are nested. But I like the clarity that you get - I don't expect many people to nest these after eachother. 

But why don't you want this grouping:
```
(3 if true else -1) * (10 if false else 2) + 2
```
I feel like the binding rules you propose are inconsistent - since they seam to have a "rest of the line" kind of semantic to them.

Why is the first 3 cut of but not the last 2? If I were the first 3 I would be pissed!

Either everyone gets a buddy - or noone gets a buddy! >:(

sornas (approved):

PR #274, opened 2021-07-04
Cleaner syntax for type constants

FredTheDino:
Type constants are now less different.

Instead of the old syntax:
```
#int | str
```
I changed the syntax to:
```
: int | str
```
This way - it motivates why the declaration-statements are the way they are:
```
a   : int | str = 3
b = : int | str
```

I feel pretty smart!


sornas (approved):
big :+1:

PR #275, opened 2021-07-04
Spans with column information in the parser and compiler

sornas:
Check it out!

```
// a.sy
start :: fn {
    // åäö <-- look ma, variable length encoding!
    abc := wow()
    a := 1
}
```
```
$ NO_COLOR=1 cargo run a.sy
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/sylt a.sy`
compile error: a.sy:3
      Failed to compile line 3
      Cannot read 'wow' in 'a.sy'
   1 | start :: fn {
   2 |     // åäö <-- look ma, variable length encoding!
   3 |     abc := wow()
                  ^^^

Error: "1 errors occured."
```

The compiler and type checker still only knows about lines though.

### DIFF
--- a/sylt-common/src/error.rs
+++ b/sylt-common/src/error.rs
@@ -8,6 +8,45 @@ use std::io::{self, BufRead};
 use std::path::{Path, PathBuf};
 use sylt_tokenizer::{Span, Token};
 
+static INDENT: &'static str = "      ";
+
+fn underline(f: &mut fmt::Formatter<'_>, col_start: usize, len: usize) -> fmt::Result {
+    write!(f, "{}", INDENT)?;
+    write!(
+        f,
+        "{: <1$}",
+        "",
+        col_start,
+    )?;
+    writeln!(
+        f,
+        "{:^<1$}",
+        "",
+        len,
+    )
+}
+
+fn write_source_line_at(f: &mut fmt::Formatter<'_>, file: &Path, span: Span) -> fmt::Result {
+    let file = if let Ok(file) = File::open(file) {
+        file
+    } else {
+        return write!(f, "Unable to open file {}", file.display());
+    };
+
+    let start_line = (span.line - 2).max(1);
+    let lines = span.line - start_line + 1;
+
+    for (line_num, line) in io::BufReader::new(file)
+        .lines()
+        .enumerate()
+        .skip(start_line - 1)
+        .take(lines)
+    {
+        writeln!(f, " {:3} | {}", (line_num + 1).blue(), line.unwrap())?;
+    }
+    underline(f, span.col_start, span.col_end - span.col_start)
+}
+
 fn source_line_at(file: &Path, line: Option<usize>) -> Option<String> {
     match (File::open(file), line) {
         (Ok(file), Some(line)) => Some(
@@ -111,14 +150,12 @@ pub enum Error {
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let indent = "      ";
-
         match self {
             #[rustfmt::skip]
             Error::RuntimeError { kind, phase, file, line, message } => {
                 write!(f, "{} {}: ", phase.red(), "error".red())?;
                 write!(f, "{}\n", file_line_display(file, Some(*line)))?;
-                write!(f, "{}{}\n", indent, kind)?;
+                write!(f, "{}{}\n", INDENT, kind)?;
 
                 if let Some(message) = message {
                     write!(f, "{}\n", message)?;
@@ -136,28 +173,13 @@ impl fmt::Display for Error {
             } => {
                 write!(f, "{}: ", "compile error".red())?;
                 write!(f, "{}\n", file_line_display(file, Some(span.line)))?;
-                write!(f, "{}Failed to compile line {}\n", indent, span.line)?;
+                write!(f, "{}Failed to compile line {}\n", INDENT, span.line)?;
 
                 if let Some(message) = message {
-                    write!(f, "{}{}\n", indent, message)?;
+                    write!(f, "{}{}\n", INDENT, message)?;
                 }
-                write!(
-                    f,
-                    "{}",
-                    source_line_at(file, Some(span.line)).unwrap_or_else(String::new)
-                )?;
-                write!(
-                    f,
-                    "{: <1$}",
-                    "",
-                    span.col_start + 6,
-                )?;
-                write!(
-                    f,
-                    "{:^<1$}\n",
-                    "",
-                    span.col_end - span.col_start,
-                )
+
+                write_source_line_at(f, file, *span)
             }
             Error::SyntaxError {
                 file,
@@ -170,11 +192,11 @@ impl fmt::Display for Error {
                 write!(
                     f,
                     "{}Syntax Error on line {} at token {:?}\n",
-                    indent, line, token
+                    INDENT, line, token
                 )?;
 
                 if let Some(message) = message {
-                    write!(f, "{}{}\n", indent, message)?;
+                    write!(f, "{}{}\n", INDENT, message)?;
                 }
                 write!(
                     f,
@@ -186,7 +208,7 @@ impl fmt::Display for Error {
                 write!(f, "{}: ", "git conflict error".red())?;
                 write!(f, "{}\n", file_line_display(file, Some(*start)))?;
 
-                write!(f, "{}Git conflict marker found at line {}\n", indent, start)?;
+                write!(f, "{}Git conflict marker found at line {}\n", INDENT, start)?;
 
                 write!(
                     f,
--- a/sylt-common/src/error.rs
+++ b/sylt-common/src/error.rs
@@ -301,10 +301,10 @@ mod test {
         }
     }
 
-    struct SourceSpanTester(super::Span);
-    impl std::fmt::Display for SourceSpanTester {
+    struct SourceSpanTester<'p>(&'p std::path::Path, super::Span);
+    impl<'p> std::fmt::Display for SourceSpanTester<'p> {
         fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-            super::write_source_span_at(f, &get_tmp(), self.0)
+            super::write_source_span_at(f, self.0, self.1)
         }
     }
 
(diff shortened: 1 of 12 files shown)


sylt-common/src/error.rs
```diff
@@ -281,3 +279,97 @@ impl fmt::Display for RuntimeError {
...
+    #[test]
+    fn underline() {
+        assert_eq!(&format!("{}", UnderlineTester(1, 2)), " ^^\n");
+        assert_eq!(&format!("{}", UnderlineTester(2, 1)), "  ^\n");
+        assert_eq!(&format!("{}", UnderlineTester(0, 0)), "\n");
+        assert_eq!(&format!("{}", UnderlineTester(0, 2)), "^^\n");
+    }
+
+    fn get_tmp() -> std::path::PathBuf {
+        use rand::Rng;
```
FredTheDino, line 316: ...
https://crates.io/crates/sungod

Do we even need rand to begin with? Why not give in a funky string that makes the files unique?

sylt-common/src/error.rs
```diff
@@ -281,3 +279,97 @@ impl fmt::Display for RuntimeError {
...
+                        col_end: 6,
+                    }
+                )
+            ),
+            "   2 | hello
+   3 | hello
+   4 | start :: fn {
+       ^^^^^\n",
+        );
+    }
```
FredTheDino, line 374: These feel macroable

sylt-compiler/src/compiler.rs
```diff
@@ -1031,8 +1031,8 @@ impl Compiler {
         }
         let module = &tree.modules[0].1;
 
-        self.read_identifier("start", Span { line: 0 }, ctx, 0);
-        self.add_op(ctx, Span { line: 0 }, Op::Call(0));
+        self.read_identifier("start", Span::zero(), ctx, 0);
+        self.add_op(ctx, Span::zero(), Op::Call(0));
```
FredTheDino, line 1035: +1

sylt-parser/src/parser.rs
```diff
@@ -6,13 +6,22 @@ use std::path::{Path, PathBuf};
...
 pub use self::statement::{Statement, StatementKind};
 
+pub use sylt_tokenizer::Span;
+
+static INVALID_TOKEN: Token = Token::EOF;
+static INVALID_SPAN: Span = Span {
+    line: 0,
+    col_start: 0,
+    col_end: 0,
+};
```
FredTheDino, line 23: This is called both `INVALID_SPAN` and `zero`-span. But they have no difference. I suspect this should be called `ZERO_SPAN`, but you already have a function for that.

sylt-parser/src/parser.rs
```diff
@@ -6,13 +6,22 @@ use std::path::{Path, PathBuf};
...
+use sylt_tokenizer::{PlacedToken, Token, file_to_tokens};
 
 pub mod expression;
 pub mod statement;
 pub use self::expression::{Expression, ExpressionKind};
 pub use self::statement::{Statement, StatementKind};
 
+pub use sylt_tokenizer::Span;
+
+static INVALID_TOKEN: Token = Token::EOF;
```
FredTheDino, line 18: This is a bit confusing. `EOF` isn't invalid. I don't think these constants add anything for readability.

sylt-parser/src/parser.rs
```diff
@@ -182,43 +182,36 @@ pub struct Type {
...
 
 /// Keeps track of where the parser is currently parsing.
 #[derive(Debug, Copy, Clone)]
 pub struct Context<'a> {
     pub skip_newlines: bool,
     /// All tokens to be parsed.
-    pub tokens: &'a Tokens,
-    /// The index of the curren token in the tokens slice.
+    pub tokens: &'a [Token],
+    /// The corresponding span for each token. Matches 1:1 with the tokens.
```
FredTheDino, line 193: Why split them out of their tuples?

sylt-common/src/error.rs
```diff
@@ -281,3 +279,97 @@ impl fmt::Display for RuntimeError {
...
+#[cfg(test)]
+mod test {
+    // A small hack is required to test the functions working on Formatters
+    // since we can't construct new Formatters.
+    //
+    // The formatted span output is very weird to test. Feel free to move the
+    // string around and check that the ^^^ lines up correctly. Spaces matter!
+    //
+    // For some tests, color is disabled by setting the env variable NO_COLOR=1.
+
```
FredTheDino, line 292: I love tests! <3

sylt-parser/src/parser.rs
```diff
@@ -6,13 +6,22 @@ use std::path::{Path, PathBuf};
...
 pub use self::expression::{Expression, ExpressionKind};
 pub use self::statement::{Statement, StatementKind};
 
+pub use sylt_tokenizer::Span;
+
+static INVALID_TOKEN: Token = Token::EOF;
+static INVALID_SPAN: Span = Span {
+    line: 0,
+    col_start: 0,
+    col_end: 0,
```
FredTheDino, line 22: Do we use only lines somewhere? Because this way to structure the information feels inferior to byte-offsets. Isn't it limiting to only have errors on one line - like if-statements and the like?

sylt-tokenizer/src/tokenizer.rs
```diff
@@ -4,42 +4,99 @@ pub use token::Token;
...
+pub struct Span {
+    // TODO(ed): Do this more intelligent, so
+    // we can show ranges. Maybe even go back
+    // to offsets from start of the file.
+    pub line: usize,
+    /// The first column that this Span contains.
+    pub col_start: usize,
+    /// The first column that this Span doesn't contain.
+    pub col_end: usize,
+}
```
FredTheDino, line 18: Haven't I seen this already? Why not move it to `common`?

sylt-tokenizer/src/tokenizer.rs
```diff
@@ -4,42 +4,99 @@ pub use token::Token;
...
+    // We also need to keep track of the current line and which char index the
+    // previous newline was at. Given a char index we can then subtract the last
+    // newline char index and get the column in the current line of the char.
+    let mut line = 1;
+    let mut last_newline = 0;
+
+    Token::lexer(&content)
+        .spanned()
+        .map(|(token, byte_range)| {
+            let is_newline = token == Token::Newline;
```
FredTheDino, line 64: Why not move this check into the if-statement where it's being used?

sylt-tokenizer/src/tokenizer.rs
```diff
@@ -53,6 +110,69 @@ mod tests {
...
+            string_to_tokens("wow\nwöw\n"),
+            (Token::Identifier(String::from("wow")), 1, 1..4),
+            (Token::Newline,                         1, 4..5),
+
+            (Token::Identifier(String::from("w")),   2, 1..2),
+            (Token::Error,                           2, 2..3),
+            (Token::Identifier(String::from("w")),   2, 3..4),
+            (Token::Newline,                         2, 4..5),
+        );
+    }
```
FredTheDino, line 174: I do love tests and macros!

FredTheDino (approved):
Very cool! Can we propegate these changes even further. So we can get span errors in the runtime and typechecker as well? :o

sylt-common/src/error.rs
sornas: The funky string has to be unique for every test and a random number felt like the easiest. But I forgot the sungod existed. All hail!

sylt-parser/src/parser.rs
sornas: The EOF can be inlined, but the other static (the zero-span) needs to be a static. Otherwise we can't return a reference to it in `peek`: https://github.com/FredTheDino/sylt-lang/blob/fecb55230a1d60d8fc7e364aafd9da6d8ba2e4cb/sylt-parser/src/parser.rs#L254

sylt-parser/src/parser.rs
sornas: At some point I think I had to, but currently it's only because I think this:
```rust
    let (ctx, kind) = match &ctx.tokens[ctx.curr..] {
        // ...
        [T::Identifier(name), T::ColonColon, ..] => {
```
is nicer to read than this:
```rust
    let (ctx, kind) = match &ctx.tokens[ctx.curr..] {
        // ...
        [(T::Identifier(name), _), (T::ColonColon, _), ..] => {
```

sylt-tokenizer/src/tokenizer.rs
sornas: `sylt-common` depends on `sylt-tokenizer`, and since it's needed in the tokenizer it now lives there.

sylt-parser/src/parser.rs
sornas: The advantage of parsing the byte offsets into lines and columns early is that this way, only the tokenizer cares about byte offsets. If we only stored the byte offsets, the error reporter would have to calculate the line and column _for each singular error_, which feels wasteful since it doesn't care at all about the byte offsets.

Storing multiple lines can be done with two fields `line_start` and `line_end`. My idea was that spans could be joined, something like this:
```diff
         [T::If, ..] => {
+            let if_span = ctx.span();
+            // if_span = Span {
+            //     line_start: 2
+            //     line_end: 2
+            //     col_start: 4
+            //     col_end: 6
+            // }
             let (ctx, condition) = expression(ctx.skip(1))?;
  
             let (ctx, pass) = statement(ctx)?;
             // else?
             let (ctx, fail) = if matches!(ctx.token(), T::Else) {
                 let (ctx, fail) = statement(ctx.skip(1))?;
                 (ctx, fail)
             } else {
                 // No else so we insert an empty statement instead.
                 (
                     ctx,
                     Statement {
                         span: ctx.span(),
                         kind: EmptyStatement,
                     },
                 )
             };
  
+            let span = if_span.join(ctx.span());
+            // span = Span {
+            //     line_start: 2
+            //     line_end: 4
+            //     col_start: 4
+            //     col_end: 5
+            // }
             (
                 ctx,
                 If {
+                    span,
                     condition,
                     pass: Box::new(pass),
                     fail: Box::new(fail),
                 },
             )
         }

```

I didn't include this here since this felt self-contained.

sylt-parser/src/parser.rs
sornas: :+1:. See https://github.com/FredTheDino/sylt-lang/pull/275#discussion_r663850706 for why the static is needed.

sylt-tokenizer/src/tokenizer.rs
sornas: `token` is moved here: https://github.com/FredTheDino/sylt-lang/blob/compiler-spans/sylt-tokenizer/src/tokenizer.rs#L80

sornas:
> Can we propegate these changes even further. So we can get span errors in the runtime and typechecker as well?

I thought about it, but in the interest of small PRs I left it out.

sylt-tokenizer/src/tokenizer.rs
sornas: What you saw was the static which has now been moved here as well.

PR #280, opened 2021-07-09
Is-statemnts

FredTheDino:
Compiletime type constraints:

```
: A is : B
```

Fails with a compile time error if the type requirements are not met.

Great for making sure things are the way you want them to be.

I broke one thing here:
```
: A is a
```
And i'm unsure how to fix it. But this is parsed as a statement - which it isn't, it should be an is-expression. You can do this though:
```
(: A is a)
```

### DIFF
--- a/progs/tests/is/faulty_checking.sy
+++ b/progs/tests/is/faulty_checking.sy
@@ -24,5 +24,5 @@ C :: blob {
 
 start :: fn { }
 
-// error: Error::CompileError { .. } 
-// error: Error::CompileError { .. } 
+// error: Error::CompileError { .. }
+// error: Error::CompileError { .. }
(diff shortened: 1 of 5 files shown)


progs/tests/is/faulty_checking.sy
```diff
@@ -0,0 +1,28 @@
...
+:B is :B
+:C is :C
+
+:B is :A
+:C is :A
+
+start :: fn { }
+
+// error: Error::CompileError { .. } 
+// error: Error::CompileError { .. }
```
sornas, line 28: Trailing spaces! :o

sornas (approved):

progs/tests/is/faulty_checking.sy
FredTheDino:
```suggestion
// error: Error::CompileError { .. }
// error: Error::CompileError { .. }
```

PR #311, opened 2021-07-28
Summer jam - split commits

sornas:


sylt-typechecker/src/typechecker.rs
```diff
@@ -1011,7 +1024,10 @@ mod op {
         match (a, b) {
             (Type::Float, Type::Float) => Type::Float,
             (Type::Int, Type::Int) => Type::Int,
-            (Type::Tuple(a), Type::Tuple(b)) if a.len() == b.len() => tuple_bin_op(a, b, div),
+            // TODO(ed): These need to be removed in more places.
+            // (Type::Tuple(a), Type::Tuple(b)) if a.len() == b.len() => tuple_bin_op(a, b, div),
```
FredTheDino, line 1028: Good idea!

FredTheDino (approved):

sylt-compiler/src/compiler.rs
```diff
@@ -683,10 +683,23 @@ impl Compiler {
...
+                .or_else(|| {
+                    error!(
+                        self,
+                        ctx,
+                        assignable.span,
+                        "While parsing namespace '{}' is not a namespace",
+                        ident.name
+                    );
                 None
+                })
```
FredTheDino, line 702: This is a live bug fix I did during the jam. It has something to do with globals.

sylt-std/src/sylt.rs
```diff
@@ -168,12 +307,64 @@ sylt_macro::extern_function!(
...
+    "Converts something to an int"
     [One(Float(t))] -> Type::Int => {
         Ok(Int(*t as i64))
     },
 );
 
+sylt_macro::extern_function!(
+    "sylt_std::sylt"
+    as_char
+    "Converts the first char in a string to an int"
```
FredTheDino, line 319: Force the string to be of length 1?

FredTheDino (approved):

FredTheDino (approved):

sornas:
So when I tried the tuple things I got into a reaal weird error. I've reverted the change here and am creating an issue so this isn't blocked. It's fine for the version thing we had in mind since this only removes behaviour (?).

sornas:
image

yikes

FredTheDino:
> @FredTheDino Do you want to merge this with -6% coverage or do we write some tests? The decrease is mostly because of FlatValue and the std additions.

Nah - it's fine. The extra 6% is hard to reach and probably not worth it. In the long term - it would be nice with tests for the networking. But I personally don't see how we can do that right now. But it should probably be an issue.

We need to do a coverage pass and try to raise the coverage after this bugfixing-spree!

PR #320, opened 2021-08-11
Missleading error

### LINKED ISSUE #301: Misleading error when missing a `,` in blob initialization
```
A :: blob {
    a: str,
    b: str,
    c: str,
    d: str,
    e: str,
}

start :: fn {
    a := A {
        a: "a",
        b: "b",
        c: "c",
        d: "d"
        e: "e",
    }
}
```
Current error output:
```
compile error: a.sy:12
      Failed to compile line 12
      Cannot read 'b' in 'a.sy'
  10 |     a := A {
  11 |         a: "a",
  12 |         b: "b",
               ^
```
Suggested error output:
```
compile error: a.sy:11
      Failed to compile line 15
      Unexpected token when parsing blob field
  13 |         c: "c",
  14 |         d: "d"
  15 |         e: "e",
               ^
```
I think this error output is possible but I haven't looked into it.

FredTheDino:
Does some minor restructuring - and fixes #301 

This is a way better error message honestly!


sylt-parser/src/statement.rs
```diff
@@ -168,8 +165,14 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
         [T::LeftBrace, ..] => match (block_statement(ctx), expression(ctx)) {
             (Ok((ctx, stmt)), _) => (ctx, stmt.kind),
             (_, Ok((ctx, value))) => (ctx, StatementExpression { value }),
-            (Err((ctx, _)), Err(_)) => {
-                raise_syntax_error!(ctx, "Neither a block nor a valid expression");
+            (Err((_, mut stmt_errs)), Err((_, mut expr_errs))) => {
+                let errs = vec![
+                    syntax_error!(ctx, "Neither a valid block nor expression - inspect the two errors bellow"),
```
sornas, line 170:
```suggestion
                    syntax_error!(ctx, "Neither a valid block nor a valid expression - check the two errors below"),
```

sylt-parser/src/parser.rs
```diff
@@ -310,6 +310,21 @@ macro_rules! expect {
     };
 }
 
+/// Eat until any one of the specified tokens or reaches EOF.
```
sornas, line 313:
```suggestion
/// Eat until any one of the specified tokens or EOF.
```

sylt-parser/src/expression.rs
```diff
@@ -744,6 +744,13 @@ fn set_or_dict<'t>(ctx: Context<'t>) -> ParseResult<'t, Expression> {
                     exprs.push(expr);
                 }
 
+                if !matches!(ctx.token(), T::Comma | T::RightBrace) {
+                    raise_syntax_error!(
+                        ctx,
+                        "Expected an element delimiter ',' - but got {:?}",
```
sornas, line 750:
```suggestion
                        "Expected an element delimiter ',' but got {:?}",
```

sornas (approved):

sylt-parser/src/expression.rs
```diff
@@ -744,6 +744,13 @@ fn set_or_dict<'t>(ctx: Context<'t>) -> ParseResult<'t, Expression> {
                     exprs.push(expr);
                 }
 
+                if !matches!(ctx.token(), T::Comma | T::RightBrace) {
+                    raise_syntax_error!(
+                        ctx,
+                        "Expected an element delimiter ',' but got {:?}",
```
EMattfolk, line 750:
```suggestion
                        "Expected a comma ',' but got {:?}",
```
It is a bit clunky to read. Is there a good reason for writing "element delimiter"?

sylt-parser/src/parser.rs
```diff
@@ -648,9 +663,7 @@ fn module(path: &Path, token_stream: &[PlacedToken]) -> (Vec<PathBuf>, Result<Mo
                 errors.append(&mut errs);
 
                 // "Error recovery"
-                while !matches!(ctx.token(), T::EOF | T::Newline) {
-                    ctx = ctx.skip(1);
-                }
+                ctx = until!(ctx, T::Newline);
                 ctx
```
EMattfolk, line 667:
```suggestion
                until!(ctx, T::Newline)
```
Or is it unclear?

sylt-parser/src/parser.rs
```diff
@@ -310,6 +310,21 @@ macro_rules! expect {
...
+            let mut ctx = $ctx;
+            while !matches!(ctx.token(), T::EOF | $( $token )|*) {
+                ctx = ctx.skip(1);
+            }
+            ctx
+        }
+    };
+}
+
+
```
EMattfolk, line 327: Couldn't this be a regular function?

EMattfolk (approved):

sylt-parser/src/parser.rs
FredTheDino: Yes - it could be.
BUT:
 - Not if you want to allow multiple tokens passed into the `matches` macro - no use of it right now so I might be over zealous
  - I also felt it was more inline with how "expect" works
  
It's also fair to mention `skip_if` which functions similarly and is infact a function.

sornas: While on the subject on this macro, maybe `skip_until` instead?

PR #322, opened 2021-08-11
End statements correctly

EMattfolk:
Tests still need to be written...


sylt-parser/src/expression.rs
```diff
@@ -169,15 +169,7 @@ fn function<'t>(ctx: Context<'t>) -> ParseResult<'t, Expression> {
     };
 
     // Parse the function statement.
-    let (ctx, mut statement) = block_statement(ctx)?;
-
-    // If the return type isn't void, check for and apply implicit returns.
```
FredTheDino, line 174: Keep this.

sylt-parser/src/expression.rs
```diff
@@ -169,15 +169,7 @@ fn function<'t>(ctx: Context<'t>) -> ParseResult<'t, Expression> {
...
-
-    let statements = if let StatementKind::Block { statements } = &mut statement.kind {
-        statements
-    } else {
-        unreachable!("Function blocks should only be blocks");
-    };
+    let (ctx, mut statements) = block(ctx)?;
 
     if !matches!(ret.kind, Resolved(Void)) {
```
FredTheDino, line 174:
```suggestion
    // If the return type isn't void, check for and apply implicit returns.
    if !matches!(ret.kind, Resolved(Void)) {
```

sylt-parser/src/parser.rs
```diff
@@ -219,6 +219,13 @@ impl<'a> Context<'a> {
         new
     }
 
+    /// Back up one token. Will not move past the beginning.
```
FredTheDino, line 222:
```suggestion
    /// Back up one token. Will not move before the beginning.
```
Is it clearer or nah?

sylt-parser/src/statement.rs
```diff
@@ -120,8 +120,7 @@ pub struct Statement {
     pub kind: StatementKind,
 }
 
-pub fn block_statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
-    let span = ctx.span();
+pub fn block<'t>(ctx: Context<'t>) -> ParseResult<'t, Vec<Statement>> {
```
FredTheDino, line 123: +1

sylt-parser/src/statement.rs
```diff
@@ -144,10 +143,10 @@ pub fn block_statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
         }
     }
 
-    let ctx = expect!(ctx, T::RightBrace, "Expected }} after block statement");
     if errs.is_empty() {
+        let ctx = expect!(ctx, T::RightBrace, "Expected }} after block statement");
```
FredTheDino, line 147:
```suggestion
        let ctx = expect!(ctx, T::RightBrace, "Expected }} after block");
```

sylt-parser/src/statement.rs
```diff
@@ -457,7 +456,13 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
     };
 
-    let ctx = ctx.skip_if(T::Newline);
+    // Newline, RightBrace and Else can end a statment.
+    // We don't give too specific errors when that is not the case.
+    let ctx = if !matches!(ctx.token(), T::RightBrace | T::Else) {
+        expect!(ctx, T::Newline, "Expected newline")
+    } else {
+        ctx
+    };
```
FredTheDino, line 465:
```suggestion
    let ctx = if matches!(ctx.token(), T::RightBrace | T::Else) {
		ctx
    } else {
        expect!(ctx, T::Newline, "Expected newline to end statement")
    };
```

sylt-parser/src/statement.rs
```diff
@@ -457,7 +456,13 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
         }
     };
 
-    let ctx = ctx.skip_if(T::Newline);
+    // Newline, RightBrace and Else can end a statment.
+    // We don't give too specific errors when that is not the case.
```
FredTheDino, line 460:
```suggestion
    // Errors only report on the newline - since it's usually the solution.
```
Maybe? I dislike the double negation in the comment - makes it hard to follow.

FredTheDino (approved):
Nice! Great work! <3

sylt-parser/src/statement.rs
sornas: This suggestion has tabs!!

sylt-parser/src/statement.rs
sornas: What's "the solution"?

sylt-parser/src/parser.rs
EMattfolk: I would say nah

sylt-parser/src/statement.rs
EMattfolk:
```suggestion
    // If a statement does not end, we only report it as a missing newline.
```
Perhaps?

PR #324, opened 2021-08-11
Better error formatting

FredTheDino:
Added a better error message - it does result in some dumb errors, but I have a sneaking suspicion they are fixed by other branches.

### DIFF
--- a/sylt-parser/src/expression.rs
+++ b/sylt-parser/src/expression.rs
@@ -320,7 +320,7 @@ fn prefix<'t>(ctx: Context<'t>) -> ParseResult<'t, Expression> {
                 )),
                 (Err((_, mut blob_errs)), Err((_, mut ass_errs))) => {
                     let errs = vec![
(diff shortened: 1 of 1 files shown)


sylt-parser/src/expression.rs
```diff
@@ -318,8 +318,13 @@ fn prefix<'t>(ctx: Context<'t>) -> ParseResult<'t, Expression> {
                         kind: Get(assign),
                     },
                 )),
-                (Err((ctx, _)), Err(_)) => {
-                    raise_syntax_error!(ctx, "Neither a blob instantiation or an identifier");
+                (Err((_, mut blob_errs)), Err((_, mut ass_errs))) => {
+                    let errs = vec![
+                        syntax_error!(ctx, "Neither a blob instantiation or an identifier - see the 2 errors below"),
```
sornas, line 323:
```suggestion
                        syntax_error!(ctx, "Neither a blob instantiation or an identifier - check the two errors below"),
```
Consistency is nice :>

sornas (approved):

EMattfolk (approved):

PR #326, opened 2021-08-11
0.2.0

sornas:
Now that #311 is merged we can think about tagging a release. This PR does two things that help with this.
1) Add a "version = 0.1.0" to all internal dependencies. This is needed in case we want to publish to crates.io and this felt like the right time. This also makes it a prime candidate for a 0.1.0-tag.
2) Bump to 0.2.0. We don't really care about semver technicalities just yet (imo) since it's easier to move all versions at the same time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -895,7 +895,7 @@ checksum = "b351cad1154eca506fab7bc868ffa732e641171a53d6e75aeaae0a042269c6f4"
 
 [[package]]
 name = "sylt"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "criterion",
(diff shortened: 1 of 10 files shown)


FredTheDino (approved):
Feels like it's annoying to bump the versions - but looks good to me!

sornas:
> Feels like it's annoying to bump the versions - but looks good to me!

`echo **/Cargo.toml | xargs sed -i "s/0.1.0/0.2.0/g"` is the super easy fix. I don't know enough sed to only match on lines containing `sylt` but I believe it could be done. Also, cargo complains if something is wrong. It could be better but it could be worse.

Issue #298, opened 2021-07-21
Declaring a tuple in a loop with the return value from an external function does not update the tuple value after the first call


sornas:
https://github.com/FredTheDino/PuckNet/commit/95f32c6e49284fc98020c910fab4b831173868b1

FredTheDino:
So this bug is surprisingly complex. The original error was that `break` and `continue` were improperly implemented in the compiler. BUT they were implemented to support an invariant - namely that no varaiable can be popped from two places. This is what breaks the typechecker and is a limitation in the way we've implemented it.

So we need something to disable the typechecker for cirtain tests to properly verify this - until we rewrite the typesystem.

PR #332, opened 2021-08-13
Fix breaks and continues in loops - but break the typechecker

FredTheDino:
- Fix breaks and continues in loops - but brake the typechecker
- Add workaround for break and continue
- Add tests for declaring variables in loops

Solves #298

### DIFF
--- a/sylt-compiler/src/compiler.rs
+++ b/sylt-compiler/src/compiler.rs
@@ -276,7 +276,7 @@ impl Compiler {
 
     fn emit_pop_until_size(&mut self, ctx: Context, span: Span, target_size: usize) {
         let vars: Vec<_> = self.frames[ctx.frame].variables.iter().skip(target_size).rev().cloned().collect();
-        for var in vars.iter() {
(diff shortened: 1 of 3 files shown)


sylt-compiler/src/compiler.rs
```diff
@@ -274,9 +274,9 @@ impl Compiler {
             .curr()
     }
 
-    fn pop_until_size(&mut self, ctx: Context, span: Span, target_size: usize) {
-        while target_size < self.frames[ctx.frame].variables.len() {
-            let var = self.frames[ctx.frame].variables.pop().unwrap();
+    fn emit_pop_until_size(&mut self, ctx: Context, span: Span, target_size: usize) {
+        let vars: Vec<_> = self.frames[ctx.frame].variables.iter().skip(target_size).rev().cloned().collect();
+        for var in vars.iter() {
```
sornas, line 279:
```suggestion
        for var in &vars {
```

sornas (approved):

PR #333, opened 2021-08-13
Less noisy errors

FredTheDino:
- Driveby grammer
- Driveby: Fix warning
- Filter out errors for seen tokens

Related to #320 - but probably not all the way there.
It's a first step anoyhow.

### DIFF
--- a/sylt-parser/src/parser.rs
+++ b/sylt-parser/src/parser.rs
@@ -822,10 +822,7 @@ pub fn tree(path: &Path) -> Result<AST, Vec<Error>> {
         let mut seen = HashSet::new();
         let errors = errors.into_iter().filter(|err| match err {
             Error::SyntaxError { span, file, .. } => {
-                let id = (span.clone(), file.clone());
-                let unseen = !seen.contains(&id);
-                seen.insert(id);
-                unseen
+                seen.insert((span.clone(), file.clone()))
             }
 
             _ => true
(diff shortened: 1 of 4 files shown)


EMattfolk (approved):

sylt-parser/src/parser.rs
```diff
@@ -818,6 +818,18 @@ pub fn tree(path: &Path) -> Result<AST, Vec<Error>> {
...
+            Error::SyntaxError { span, file, .. } => {
+                let id = (span.clone(), file.clone());
+                let unseen = !seen.contains(&id);
+                seen.insert(id);
+                unseen
+            }
+
+            _ => true
+        }).collect();
         Err(errors)
```
sornas, line 833: You could inline `errors`. Also, filter with side-effects!! :O

sylt-parser/src/parser.rs
```diff
@@ -818,6 +818,18 @@ pub fn tree(path: &Path) -> Result<AST, Vec<Error>> {
...
         Ok(AST { modules })
     } else {
+        // Filter out errors for already seen spans
+        let mut seen = HashSet::new();
+        let errors = errors.into_iter().filter(|err| match err {
+            Error::SyntaxError { span, file, .. } => {
+                let id = (span.clone(), file.clone());
+                let unseen = !seen.contains(&id);
+                seen.insert(id);
+                unseen
```
sornas, line 828:
```suggestion
                seen.insert(id).is_none()
```
;)

But maybe it's a bit too unclear what actually happens then.

sornas (approved):

PR #334, opened 2021-08-13
Don't write values with 'dbg' in the typechecker

### LINKED ISSUE #287: Don't write values in `dbg` if typechecking - it confuses people

FredTheDino:
Make it noob friendly! :D

Closes #287


EMattfolk (approved):
LGTM, but I think the title of the pull request should be updated. Is it missing "in" before "the"?

PR #338, opened 2021-08-14
Better prints

### LINKED ISSUE #188: Better prints
This should be a function, taking mutliple arguments. Think python.

Handling recursive objects would be nice but not nessecary.

EMattfolk:
Values should look nice when printed, therefore we made them look good when printed.

We also handle recursive values:
```
a := []
a -> push' a
print a
// [[...]]
```

The only thing missing is a print function.
Closes #188


sylt-common/src/value.rs
```diff
@@ -172,6 +160,115 @@ impl Value {
             Value::ExternFunction(slot) => slot + 2,
         }
     }
+
+    fn safe_fmt(
```
sornas, line 164: small doc here

sylt-common/src/value.rs
```diff
@@ -172,6 +160,115 @@ impl Value {
...
+        match self {
+            Value::Field(s) => write!(fmt, "<field .{}>", s),
+            Value::Ty(ty) => write!(fmt, "<type \"{:?}\">", ty),
+            Value::Blob(b) => write!(fmt, "<blob \"{}\">", b),
+            Value::Instance(_, v) => {
+                write!(fmt, "{} {{",
+                    if let Some(Value::String(name)) = v.borrow().get("_name") {
+                        name.as_str()
+                    } else {
+                        "?"
```
sornas, line 178: Maybe the memory location? At least _something_ that is unique per instance.

sylt-common/src/value.rs
```diff
@@ -172,6 +160,115 @@ impl Value {
...
+                        continue;
+                    }
+                    if !first {
+                        write!(fmt, ", ")?;
+                    }
+                    write!(fmt, "{}", e.0)?;
+                    write!(fmt, ": ")?;
+                    e.1.safe_fmt(fmt, seen)?;
+                    first = false;
+                }
```
sornas, line 196: I _think_ this can be an iterator that filtermaps to `String` and uses https://doc.rust-lang.org/std/primitive.slice.html#method.join.

sornas: It would use `format!` which explicitly allocates but IIRC `write!` already allocates.

sornas: `join` would remove the trailing `, ` as well.

sylt-common/src/value.rs
```diff
@@ -172,6 +160,115 @@ impl Value {
...
+                if !seen.insert(self.unique_id()) {
+                    return write!(fmt, "[...]");
+                }
+                write!(fmt, "[")?;
+                for (i, e) in v.borrow().iter().enumerate() {
+                    if i != 0 {
+                        write!(fmt, ", ")?;
+                    }
+                    e.safe_fmt(fmt, seen)?;
+                }
```
sornas, line 216: As earlier, I think this can be a filtermap and `join`.

sylt-common/src/value.rs
```diff
@@ -172,6 +160,115 @@ impl Value {
...
+                write!(fmt, "]")
+            },
+            Value::Tuple(v) => {
+                write!(fmt, "(")?;
+                for (i, e) in v.iter().enumerate() {
+                    if i != 0 {
+                        write!(fmt, ", ")?;
+                    }
+                    e.safe_fmt(fmt, seen)?;
+                }
```
sornas, line 226: Potential filtermap and `join`.

sylt-common/src/value.rs
```diff
@@ -172,6 +160,115 @@ impl Value {
...
+                write!(fmt, "{{")?;
+                for (i, e) in v.borrow().iter().enumerate() {
+                    if i != 0 {
+                        write!(fmt, ", ")?;
+                    }
+                    e.0.safe_fmt(fmt, seen)?;
+                    write!(fmt, ": ")?;
+                    e.1.safe_fmt(fmt, seen)?;
+                }
```
sornas, line 257: dito

sylt-std/src/sylt.rs
```diff
@@ -379,8 +379,8 @@ sylt_macro::extern_function!(
     "sylt_std::sylt"
     as_str
     "Converts to a string representation"
-    [One(Int(i))] -> Type::String => {
-        Ok(Value::String(Rc::new(i.to_string())))
+    [One(v)] -> Type::String => {
+        Ok(Value::String(Rc::new(v.to_string())))
```
sornas, line 383: :+1:

progs/tests/string_conversion.sy
```diff
@@ -0,0 +1,67 @@
...
+    // List
+    a := []
+    a -> push' a
+    a -> push' 1
+    as_str(a) <=> "[[...], 1]"
+
+    // Tuple
+    print ()
+    print (1,)
+    print (1, 2)
```
FredTheDino, line 29: Prints aren't really checkable - so these only add coverage of lines we don't check.

progs/tests/string_conversion.sy
```diff
@@ -0,0 +1,67 @@
...
+
+    // Set
+    as_str({}) <=> "{}"
+    as_str({1}) <=> "{1}"
+    print {1, 2}
```
FredTheDino, line 34: *

progs/tests/string_conversion.sy
```diff
@@ -0,0 +1,67 @@
...
+    c :: {}
+    c -> add' c
+    as_str(c) <=> "{{...}}"
+
+
+    // Dict
+    as_str({:}) <=> "{:}"
+    print {1: 2}
+    print {1: 2, 3: 4}
```
FredTheDino, line 43: *

progs/tests/string_conversion.sy
```diff
@@ -0,0 +1,67 @@
...
+
+    // Blobs
+    print A
+    print A { a: 100, b: 5.3 }
```
FredTheDino, line 47: *

progs/tests/string_conversion.sy
```diff
@@ -0,0 +1,67 @@
...
+    as_str(b) <=> "B {self: B {...}}"
+
+    b._name = nil // :O
+    as_str(b) <=> "? {self: ? {...}}"
+
+    // Functions
+    d :: fn x: int, y: int -> int {
+        x + y
+    }
+    print d
```
FredTheDino, line 61: *

progs/tests/string_conversion.sy
```diff
@@ -0,0 +1,67 @@
...
+
+    // Other
+    print pop
+    print :str
```
FredTheDino, line 65: *

sylt-common/src/value.rs
FredTheDino: I think we tried this - but I think the borrow checker gets sad about `fmt`

sylt-common/src/value.rs
FredTheDino: Currently - the "?" should be unreachable. Since we don't have such a thing as anonymous blobs. But maybe it's worth adding the pointer to differentiate instances of object - like Gustav suggests. Maybe it should be written after the type?
```
A (0xdeadbeef) { .a = 1 }
```
?

sylt-std/src/sylt.rs
FredTheDino: This is what is hindering tuples to be passed in. Let me know if you want help with writing this without the macro - can be hard to do if you haven't done it before.

FredTheDino (approved):
LGTM - just need to solve Gustavs comment about separating instances and maybe the tuple conundrum.

sylt-common/src/value.rs
FredTheDino: Write unreachable or unimplemented

FredTheDino (approved):
LGTM

sornas (approved):

PR #344, opened 2021-08-17
Add a basic formatter

sornas:


sylt-tokenizer/src/token.rs
```diff
@@ -143,8 +143,8 @@ pub enum Token {
     #[token(">>>>>>>")]
     GitConflictEnd,
 
-    #[regex(r"//[^\n]*", logos::skip)]
-    Comment,
+    #[regex(r"//[^\n]*", |lex| lex.slice().to_string())]
```
FredTheDino, line 146: Should we strip the `//`?

sylt-parser/src/parser.rs
```diff
@@ -270,6 +287,17 @@ impl<'a> Context<'a> {
         &self.peek().0
     }
 
+    fn tokens_forward<const N: usize>(&self) -> [Token; N] {
+        const ERROR: Token = Token::Error;
+        let mut res = [ERROR; N];
+        let mut ctx = *self;
+        for i in 0..N {
+            res[i] = ctx.token().clone();
```
FredTheDino, line 317: You should be able to borrow them - no?

sylt-parser/src/parser.rs
```diff
@@ -270,6 +287,17 @@ impl<'a> Context<'a> {
         &self.peek().0
     }
 
+    fn tokens_forward<const N: usize>(&self) -> [Token; N] {
+        const ERROR: Token = Token::Error;
+        let mut res = [ERROR; N];
```
FredTheDino, line 314: :eyes:

sylt-parser/src/parser.rs
sornas: Not trivially. Feel free to try it.

sylt-parser/src/statement.rs
```diff
@@ -19,7 +28,8 @@ pub enum StatementKind {
     /// `use <file> as <alias>`.
     /// `use <folder>/ as <alias>`.
     Use {
-        ident: Identifier,
+        name: Identifier,
+        alias: UseIdentifier,
```
sornas, line 32: These names are hard.

sylt-parser/src/statement.rs
```diff
@@ -243,16 +262,20 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
-                    (ctx, Identifier { span, name })
+                    (ctx, UseIdentifier::Implicit(Identifier { span, name }))
                 },
             };
-            (ctx, Use { ident, file })
+            let name = Identifier {
+                span: name_span,
+                name,
+            };
+            (ctx, Use { name, alias, file })
```
sornas, line 278: I think this is correct but I'd appreciate a look at it.

sylt-parser/src/parser.rs
```diff
@@ -275,6 +319,17 @@ impl<'a> Context<'a> {
         &self.peek().0
     }
 
+    fn tokens_forward<const N: usize>(&self) -> [Token; N] {
```
sornas, line 322: This should be used when lookahead is needed. Sadly even if we don't use this it will still work when there aren't any comments.

That is, we want to do this: https://github.com/FredTheDino/sylt-lang/pull/344/commits/a4cb50e06fbae803dd282d783aa9ac65272524f2

sornas: Maybe name it `tokens_lookahead` instead?

FredTheDino: Go for it - but do it in another PR

sornas: I mean, it was introduced in this PR.

sornas:
I don't get the CI error locally :thinking:

sylt-parser/src/parser.rs
```diff
@@ -215,12 +221,39 @@ impl<'a> Context<'a> {
...
+        // Skip n tokens.
+        while skipped < n {
+            if matches!(new.token(), T::Comment(_)) {
+                // Skip comments without counting as a token.
+                new.curr += 1;
+            } else {
+                new.curr += 1;
+                skipped += 1;
+            }
+        }
```
EMattfolk, line 249:
```suggestion
        // Skip n non comment tokens.
        while skipped < n {
            if !matches!(new.token(), T::Comment(_)) {
                skipped += 1;
            }
            new.curr += 1;
        }
```

sylt-parser/src/parser.rs
```diff
@@ -186,12 +186,17 @@ type ParseResult<'t, T> = Result<(Context<'t>, T), (Context<'t>, Vec<Error>)>;
 #[derive(Debug, Copy, Clone)]
 pub struct Context<'a> {
     pub skip_newlines: bool,
+    /// The index of the last token of the last statement parsed.
```
FredTheDino, line 189:
```suggestion
    /// The index of the end token of the last statement parsed.
```
Is this clearer?

sylt-parser/src/parser.rs
FredTheDino: Then go for it!

sylt-parser/src/statement.rs
```diff
@@ -217,18 +236,18 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
-                    format!("{}/exports.sy", bare_name)
+                    format!("{}/exports.sy", name)
                 } else {
-                    format!("{}.sy", bare_name)
+                    format!("{}.sy", name)
                 })
             };
-            let (ctx, ident) = match &ctx.tokens[ctx.curr..] {
+            let (ctx, alias) = match &ctx.tokens_forward::<2>() {
                 [T::As, T::Identifier(alias), ..] => (
```
EMattfolk, line 245:
```suggestion
                [T::As, T::Identifier(alias)] => (
```

sylt-parser/src/statement.rs
```diff
@@ -217,18 +236,18 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
                     ctx.skip(2),
-                    Identifier {
+                    UseIdentifier::Alias(Identifier {
                         span: ctx.skip(1).span(),
                         name: alias.clone(),
-                    }
+                    })
                 ),
                 [T::As, ..] => raise_syntax_error!(
```
EMattfolk, line 252:
```suggestion
                [T::As, _] => raise_syntax_error!(
```

sylt-parser/src/statement.rs
```diff
@@ -1,5 +1,14 @@
 use super::*;
 
+/// The different ways a namespace is introduced by a use statement.
+#[derive(Debug, Clone)]
+pub enum UseIdentifier {
```
FredTheDino, line 5:
```suggestion
pub enum NameIdentifier {
```

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+static INDENT: &'static str = "    ";
+
+macro_rules! write_comma_separated {
+    ($dest:expr, $indent:expr, $write:expr, $values:expr) => {
+        let mut first = true;
+        for value in $values {
+            if !first {
+                write!($dest, ", ")?;
+            }
+            first = false;
```
FredTheDino, line 21: This could be an enum? So we don't have to mutate stuff?

sylt-parser/src/statement.rs
```diff
@@ -243,16 +262,20 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
                         .unwrap()
                         .to_string();
-                    (ctx, Identifier { span, name })
+                    (ctx, UseIdentifier::Implicit(Identifier { span, name }))
                 },
             };
-            (ctx, Use { ident, file })
+            let name = Identifier {
+                span: name_span,
+                name,
```
EMattfolk, line 276: Doesn't this name need to keep the slashes at the start and end? I think this might strip them.

sylt-parser/src/statement.rs
EMattfolk: I would probably do path and name instead of name and alias

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+            while let Some(expr) = exprs.next() {
+                if !first {
+                    write!(dest, ", ")?;
+                }
+                first = false;
+                write_expression(dest, indent, expr)?;
+                write!(dest, ": ")?;
+                write_expression(dest, indent, exprs.next().unwrap())?;
+            }
+            write!(dest, "}}")?;
```
FredTheDino, line 312: Is this wrong for empty dict `{:}`?

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+        ExpressionKind::Bool(b) => write!(dest, "{}", if *b { "true" } else { "false" })?,
+        ExpressionKind::Nil => write!(dest, "nil")?,
+    }
+
+    Ok(())
+}
+
+fn write_statement<W: Write>(dest: &mut W, indent: u32, statement: &Statement) -> fmt::Result {
+    for comment in &statement.comments {
+        write!(dest, "// {}\n", comment)?;
```
FredTheDino, line 326: Strip the leading and trailing whitespace?

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+                    Op::Sub => "-",
+                    Op::Mul => "*",
+                    Op::Div => "/",
+                }
+            )?;
+            write_expression(dest, indent, value)?;
+        }
+        StatementKind::Blob { name, fields } => {
+            write!(dest, "{}", name)?;
+            write!(dest, " :: blob {{\n")?;
```
FredTheDino, line 352: Why not 1 write?

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+fn write_runtime_type<W: Write>(dest: &mut W, ty: &RuntimeType) -> fmt::Result {
+    match ty {
+        RuntimeType::Void => write!(dest, "nil"),
+        RuntimeType::Int => write!(dest, "int"),
+        RuntimeType::Float => write!(dest, "float"),
+        RuntimeType::Bool => write!(dest, "bool"),
+        RuntimeType::String => write!(dest, "str"),
+        t => unreachable!("{:?} should not be present in the syntax tree", t),
+    }
+}
```
EMattfolk, line 79: This could probably be replaced now. I would double check that the Void type looks correct though.

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+                write_type(dest, indent, ty)?;
+                write!(dest, ",\n")?;
+            }
+            write_indents(dest, indent)?;
+            write!(dest, "}}")?;
+        }
+        StatementKind::Block { statements } => {
+            write!(dest, "{{\n")?;
+
+            for s in &merge_empty_statements(statements.clone()) {
```
FredTheDino, line 365: Why not make a custom iterator that skips the empty-statements?

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+                        VarKind::Const => " :: ",
+                        VarKind::Mutable => " := ",
+                        VarKind::ForceConst => unreachable!("can't force an implied type"),
+                        VarKind::ForceMutable => unreachable!("can't force an implied type"),
+                    }
+                )?;
+            } else {
+                write!(dest, ": ")?;
+                match kind {
+                    VarKind::ForceConst | VarKind::ForceMutable => write!(dest, "!")?,
```
FredTheDino, line 397: `if kind.force() { .. }`

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+                    VarKind::Const | VarKind::Mutable => (),
+                }
+                write_type(dest, indent, ty)?;
+                match kind {
+                    VarKind::Const | VarKind::ForceConst => write!(dest, " : ")?,
+                    VarKind::Mutable | VarKind::ForceMutable => write!(dest, " = ")?,
```
FredTheDino, line 403: `if kind.mutable() { .. } else ..`?

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+
+            write!(dest, "if ")?;
+            write_expression(dest, indent, condition)?;
+            write!(dest, " ")?;
+            write_statement(dest, indent, pass)?;
+            if !matches!(fail.kind, StatementKind::EmptyStatement) {
+                write!(dest, " else ")?;
+                write_statement(dest, indent, fail)?;
+            } else {
+
```
FredTheDino, line 429:
```suggestion
```

sylt/src/formatter.rs
```diff
@@ -0,0 +1,504 @@
...
+        StatementKind::Loop { condition, body } => {
+            write!(dest, "loop ")?;
+            write_expression(dest, indent, condition)?;
+            write!(dest, " ")?;
+            write_statement(dest, indent, body)?;
+        }
+        StatementKind::Ret { value } => {
+            write!(dest, "ret ")?;
+            write_expression(dest, indent, value)?;
+        }
```
FredTheDino, line 446: This won't write implicit returns - add a todo?

sylt-parser/src/statement.rs
sornas: I think you're right, yes. Will try and maybe fix.

sylt/src/formatter.rs
sornas: Why would that not need to mutate?

sylt/src/formatter.rs
sornas: double :+1:

sylt/src/formatter.rs
sornas: It is, yes.

sylt/src/formatter.rs
sornas: Leading whitespace is stripped by the tokenizer but I'll add trim_end as well.

sylt/src/formatter.rs
sornas: Interesting...

sylt/src/formatter.rs
sornas: I want to collect all todos in the same place, not spread out as comments.

sylt-parser/src/statement.rs
sornas: Disagree. `..` means that we can change the lookahead to 3 (or something else) without needing to change every branch.

sylt-parser/src/parser.rs
FredTheDino: I have also given up

sylt/src/formatter.rs
sornas: Ah well, the borrows get kinda sad. We'd either need a `Cow` or rewrite so the entire formatter moves instead of borrows. I think the former would be too much work since we _probably_ want the latter anyway.

sornas: The reason the borrows get sad is because we need to take the comments from the removed statements.

sylt/src/formatter.rs
```diff
@@ -0,0 +1,613 @@
...
+    }
+}
+
+macro_rules! expr_binary_op {
+    ($dest:expr, $indent:expr, $lhs:expr, $op:literal, $rhs:expr) => {
+        write_expression($dest, $indent, $lhs)?;
+        write!($dest, $op)?;
+        write_expression($dest, $indent, $rhs)?;
+    };
+}
```
FredTheDino, line 164: Move this down so it's above `write_expression`?

sylt/src/formatter.rs
```diff
@@ -0,0 +1,613 @@
...
+                    write!(dest, ": ")?;
+                    write_expression(dest, indent, exprs.next().unwrap())?;
+                }
+            }
+            write!(dest, "}}")?;
+        }
+        ExpressionKind::Float(f) => write!(dest, "{:?}", f)?,
+        ExpressionKind::Int(i) => write!(dest, "{}", i)?,
+        ExpressionKind::Str(s) => write!(dest, "\"{}\"", s)?,
+        ExpressionKind::Bool(b) => write!(dest, "{}", if *b { "true" } else { "false" })?,
```
FredTheDino, line 369:
```suggestion
        ExpressionKind::Bool(b) => write!(dest, if *b { "true" } else { "false" })?,
```

sylt/src/formatter.rs
```diff
@@ -0,0 +1,613 @@
...
+    // - we always want to remove and look at the first statement and
+    // - pop() is faster than remove(0).
+    statements.reverse();
+
+    let mut ret = Vec::new();
+    while let Some(mut statement) = statements.pop() {
+        if !matches!(statement.kind, StatementKind::EmptyStatement) {
+            ret.push(statement);
+            continue;
+        }
```
FredTheDino, line 532: Can't you just remove this block? The while loop below does the same check, no? I haven't tried it though.

sylt/src/formatter.rs
```diff
@@ -0,0 +1,613 @@
...
+
+                // Format the file.
+                match $crate::formatter::format(&args) {
+                    Ok(formatted) => {
+                        // Overwrite with the formatted output.
+                        ::std::fs::write(&path, formatted).unwrap();
+
+                        // Try to run the file again, this time with pretty "got/expected"-output.
+                        let after = $crate::run_file(&args, ::sylt_std::sylt::_sylt_link());
+                        eprintln!("If you can read this, the test output changed between before and after formatting");
```
FredTheDino, line 598:
```suggestion
                        eprintln!("The test output changed between before and after formatting");
```
?

sylt/src/formatter.rs
```diff
@@ -0,0 +1,613 @@
...
+                        $crate::assert_errs!(after, $errs);
+                    }
+                    Err(errs) => {
+                        eprintln!("If you can read this, the formatter couldn't parse the file but the syntax errors");
```
FredTheDino, line 602:
```suggestion
                        eprintln!("The formatter couldn't parse the file but the syntax errors");
```
?

FredTheDino (approved):
LGTM - some minor things, things that always pop up when you merge a lot of code.

sylt/src/formatter.rs
sornas:
```
error: format argument must be a string literal
   --> sylt/src/formatter.rs:355:49
    |
355 |         ExpressionKind::Bool(b) => write!(dest, if *b { "true" } else { "false" })?,
    |                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: you might be missing a string literal to format with
    |
355 |         ExpressionKind::Bool(b) => write!(dest, "{}", if *b { "true" } else { "false" })?,
    |                                                 ^^^^^
```

EMattfolk: Om det är en bool
```suggestion
        ExpressionKind::Bool(b) => write!(dest, "{}", *b)?,
```

FredTheDino (approved):

EMattfolk (approved):

PR #345, opened 2021-08-18
Don't mutate fields you're not supposed to

FredTheDino:


FredTheDino:
If people want to merge or change the error message feel free to do so - no need to ask me.

EMattfolk (commented):
I kind of feel like the name and id should not be part of the instance in the first place. At the very least not bundled with the other data in the blob.

sylt-common/src/error.rs
```diff
@@ -251,6 +252,9 @@ impl fmt::Display for RuntimeError {
             RuntimeError::UnknownField(obj, field) => {
                 write!(f, "Cannot find field '{}' on blob {:?}", field, obj)
             }
+            RuntimeError::ImmutableField(field) => {
+                write!(f, "Cannot mutate field '{}' - since it is immutable", field)
```
sornas, line 256:
```suggestion
                write!(f, "Cannot mutate field '{}' since it is immutable", field)
```

FredTheDino:
> I kind of feel like the name and id should not be part of the instance in the first place. At the very least not bundled with the other data in the blob.

Fight me!

PR #347, opened 2021-08-20
Use files in directories

### LINKED ISSUE #130: Use files in subdirectories
```
use a/b

start :: fn {
  print b.c()
}
```

- [x] `use a/b` includes file `a/b.sy` under namespace b
- [x] `use a/b/c as y` includes file `a/b/c.sy` under namespace y
- [x] `use folder/` includes file `use folder/<something>.sy`
- [x] `use /a` includes `a.sy` in the project root (folder where main.sy is)

EMattfolk:
Imports from subdirectories and renaming.


sylt-parser/src/statement.rs
```diff
@@ -168,16 +191,29 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
+        // `use path/to/file`
+        // `use path/to/file as alias`
+        [T::Use, ..] => {
+            let (path_ctx, path) = path(ctx.skip(1))?;
+            let (ctx, alias) = match &path_ctx.tokens[path_ctx.curr..] {
+                [T::As, T::Identifier(alias), ..] => (path_ctx.skip(2), Some(alias.clone())),
+                [T::As, ..] => return Err((
+                    skip_until!(path_ctx, T::Newline),
+                    vec![syntax_error!(path_ctx.skip(1), "Expected alias")]
+                )),
```
FredTheDino, line 203: Preferably use `raise_syntax_error`
```suggestion
                [T::As, ..] => raise_syntax_error(
                    skip_until!(path_ctx, T::Newline),
                    "Expected alias"
                ),
```
I haven't checked if this compiles.

sylt-parser/src/statement.rs
```diff
@@ -113,6 +116,26 @@ pub struct Statement {
...
+        result.push_str(f);
+        ctx = ctx.skip(1);
+        if matches!(ctx.token(), T::Slash) {
+            result.push_str("/");
+            ctx = ctx.skip(1);
+        }
+    }
+
+    Ok(( ctx, result ))
+}
```
FredTheDino, line 137: Neato

sylt-parser/src/statement.rs
```diff
@@ -113,6 +116,26 @@ pub struct Statement {
     pub kind: StatementKind,
 }
 
+pub fn path<'t>(ctx: Context<'t>) -> ParseResult<'t, String> {
+    let mut ctx = ctx;
+    let mut result = String::new();
+    expect!(ctx, T::Slash | T::Identifier(_), "Expected file path");
```
FredTheDino, line 122:
```suggestion
    expect!(ctx, T::Slash | T::Identifier(_), "Expected identifier or slash at start of use path");
```
Maybe something like this? I'm also not sure I like the word `file path` - but that's easily changed later.

FredTheDino (approved):
LGTM

sylt-parser/src/statement.rs
```diff
@@ -12,8 +12,11 @@ pub enum StatementKind {
     /// "Imports" another file.
     ///
     /// `use <file>`.
+    /// `use <folder>/<file>`.
+    /// `use <file> as <alias>`.
     Use {
         file: Identifier,
+        file_alias: Option<String>,
```
sornas, line 19: Wouldn't this be an `Option<Identifier>`?

sornas (approved):
What about `use /a.sy`?

EMattfolk:
> What about `use /a.sy`?

It will be added in another pr.

sylt-parser/src/statement.rs
EMattfolk: I don't think so? The point of having the file as an identifier I think is to be able to create correct errors if the file is missing. Otherwise both of them would just be strings. Maybe @FredTheDino could clear this up.

FredTheDino: What it could be - is `Vec<Identifier>` but I think the string solution should work equally well.

PR #350, opened 2021-08-21
Tree typechecker

FredTheDino:
A bunch of work - and not yet finished. But this is getting to be quite a lot of code. So if someone wants to skim it before it's finished to keep up to date - that's fine! :D


sylt-compiler/src/typechecker.rs
```diff
@@ -0,0 +1,717 @@
...
+                    return Err(vec![err]);
+                }
+                Type::Bool
+            }
+
+            EK::Neg(a) => self.uni_op(span, a, op::neg, "Negation")?,
+
+            EK::And(a, b) => self.bin_op(span, a, b, op::and, "Boolean and")?,
+            EK::Or(a, b) => self.bin_op(span, a, b, op::or, "Boolean or")?,
+            EK::Not(a) => self.uni_op(span, a, op::not, "Boolean not")?,
```
EMattfolk, line 217: I would call these logical and, or and not. But maybe they're only defined for booleans.

FredTheDino: They're only defined for booleans - and the function names match `common::Op::{And,Or,Not}` and `machine::op::{and,or,not}`. So I'd rather not change them...

FredTheDino:
There's one problem with this - and that is external functions. We now don't have a VM that can run stuff. I'm thinking we change how external functions work and maybe introduce typehints on them. That way we can test what we're documenting. :D

sylt-compiler/src/compiler.rs
```diff
@@ -1191,6 +1202,7 @@ impl Compiler {
         }
     }
 
+    // TODO(ed): This should probably be cleaned up and moves out of the compiler?
```
EMattfolk, line 1205: Yeah, or at least something has to be done to it. What, I think we should discuss.

EMattfolk (commented):
I see nothing blatantly wrong with the code, but I have yet to look at the new typechecker. I would rename the typeing folder to typing.

sornas (approved):
Looks cool. Guess we'll see how it works when we start to use it.

sylt-common/src/error.rs
```diff
@@ -282,6 +363,57 @@ impl fmt::Display for RuntimeError {
...
+                write!(f, "{} is not defined for '{:?}' and '{:?}'", op, lhs, rhs)
+            }
+
+            TypeError::UniOp { op, val } => {
+                write!(f, "{} is not defined for '{:?}'", op, val)
+            }
+
+            TypeError::Violating(ty) => {
+                write!(f, "Got '{:?}', which doesn't hold", ty)
+            }
```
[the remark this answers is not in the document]
FredTheDino: It's a very general error - we got a type that it cannot be. Feel free to come with suggestions.

PR #358, opened 2021-08-24
Fix assert precedence

### LINKED ISSUE #355: Assert precedence
The assert operator is evaluated in the wrong order. It binds too tightly.
This example does not work:

```
2 <=> fn -> int { 2 }()
```

EMattfolk:
Turns out, it was not the assert precedence that was wrong. It was
accesses with the dot and call operators that simply did not have a
precedence.

Closes #355


FredTheDino (approved):
Sicko!

PR #369, opened 2021-08-30
Fix all tests

FredTheDino:


tests/core/string_conversion.sy
```diff
@@ -59,6 +59,8 @@ start :: fn {
     print(d)
 
     // Other
-    print(A, pop, :str)
+    print' A
+    print' pop
+    print' :str
     as_str(nil) <=> "nil"
```
FredTheDino, line 65: This makes me sad...

sylt-common/src/ty.rs
```diff
@@ -214,6 +214,14 @@ impl Type {
                         ));
                     }
                 }
+                if a_args.len() != b_args.len() {
+                    return Err(
+                        format!(
+                            "mismatching arity, {} != {}",
```
EMattfolk, line 220:
```suggestion
                            "Mismatching arity, {} != {}",
```

tests/core/string_conversion.sy
EMattfolk: Not being able to print all at the same time? We could assume that they will look a specific way. Or just do a as_str instead.

tests/for_each/for_in_for.sy
```diff
@@ -2,7 +2,7 @@ start :: fn {
     values := [1,2,3,4]
     values -> for_each' fn a:int {
         values -> for_each' fn b:int {
-            print (a, b)
```
EMattfolk, line 5: 🤔

EMattfolk (approved):
Looks good. I think it is time to rename the folder to `typing` though.

sornas (approved):

tests/core/string_conversion.sy
FredTheDino: This is equivalent to what was there before - these changes should be done in a different pr.

sylt-common/src/ty.rs
FredTheDino: This is already uniform with other errors

PR #370, opened 2021-08-30
Force start to be a function

### LINKED ISSUE #348: Invalid program: `start :: {}`
```
start :: {}
```

FredTheDino:
Closes #348

### DIFF
--- a/sylt-compiler/src/typechecker.rs
+++ b/sylt-compiler/src/typechecker.rs
@@ -959,7 +959,7 @@ impl<'c> TypeChecker<'c> {
                         Ok(None) => {}
                         Err(mut errs) => {
                             errors.append(&mut errs);
-                        }
+                       }
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


sylt-compiler/src/typechecker.rs
```diff
@@ -1074,6 +1073,21 @@ impl<'c> TypeChecker<'c> {
             }
         }
 
+        self.assignable(&Assignable {
```
EMattfolk, line 1076: Maybe a short comment here?

EMattfolk (approved):

sylt-compiler/src/typechecker.rs
```diff
@@ -959,7 +958,7 @@ impl<'c> TypeChecker<'c> {
                         Ok(None) => {}
                         Err(mut errs) => {
                             errors.append(&mut errs);
-                        }
+                       }
```
sornas, line 961: :thinking:

sornas (commented):
```
gustav@arch ~/d/sylt (invalid-start)> cat a.sy
start :: {}
gustav@arch ~/d/sylt (invalid-start)> cargo run a.sy
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/sylt a.sy`
typecheck error: a.sy:0
      Got 'Set(Unknown)', which it cannot be
      Set(Unknown) cannot be called as a function


Error: "1 errors occured."
```
Can the error point to the start function? Otherwise I think the error text is too general to be usable.

PR #382, opened 2021-09-04
move instead of borrow when formatting

sornas:
This removes a couple of clones which is always nice.

Also, if we ever need to move back to borrows, this single commit is (hopefully) easily revert-able.


FredTheDino (approved):
lgtm

EMattfolk (approved):

FredTheDino:
@sornas poke

PR #386, opened 2021-09-04
Better syntax tree printing

FredTheDino:
Next time, we do a macro.

### DIFF
--- a/sylt-parser/src/expression.rs
+++ b/sylt-parser/src/expression.rs
@@ -906,7 +906,7 @@ mod test {
 impl PrettyPrint for Expression {
     fn pretty_print(&self, f: &mut std::fmt::Formatter<'_>, indent: usize) -> std::fmt::Result {
         use ExpressionKind as EK;
-        write_indent(f, indent);
+        write_indent(f, indent)?;
         match &self.kind {
             EK::Get(e) => {
                 write!(f, "Get ")?;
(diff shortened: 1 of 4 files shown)


sylt-parser/src/parser.rs
```diff
@@ -1012,3 +1012,205 @@ mod test {
...
+            TypeKind::Implied => {
+                write!(f, "Implied");
+            }
+            TypeKind::Resolved(ty) => {
+                write!(f, "{}", ty);
+            }
+            TypeKind::UserDefined(name) => {
+                write!(f, "User(")?;
+                name.pretty_print(f, 0)?;
+            }
```
sornas, line 1138: Missing `)`

sylt-parser/src/expression.rs
```diff
@@ -902,3 +902,151 @@ mod test {
...
+                    write_indent(f, indent)?;
+                    write!(f, ".{}:\n", field)?;
+                    value.pretty_print(f, indent + 1);
+                }
+            }
+            EK::Tuple(values) => {
+                write!(f, "Tuple\n")?;
+                for value in values.iter() {
+                    value.pretty_print(f, indent + 1);
+                }
```
EMattfolk, line 1023:
```suggestion
                values.iter().try_for_each(|value| value.pretty_print(f, indent + 1))
```
I think this works. It could probably be applied at other places in the repo.

EMattfolk (approved):
The -4% coverage kind of hurts. We should probably test the --dump-tree functionality.

sylt-parser/src/parser.rs
```diff
@@ -1012,3 +1012,206 @@ mod test {
...
+            TypeKind::Dict(k, v) => {
+                write!(f, "{{{}:{}}}", k, v)?;
+            }
+            TypeKind::Generic(name) => {
+                write!(f, "#{}", name.name)?;
+            }
+        }
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
FredTheDino: They can't, they're not the same kind of value. Only Resolved(ty) hold a common::Type, and they use Display.

sylt-parser/src/expression.rs
FredTheDino: Very cool

PR #387, opened 2021-09-05
Fix all warnings

FredTheDino:
Fix all warnings and make the pipeline fail if we don't fix all of them


EMattfolk (approved):

sornas (approved):
:+1: for the workflow!

PR #388, opened 2021-09-05
Create codecov.yml

FredTheDino:
This should make it informational instead of an error


sornas (approved):
A bit hard to test I guess

EMattfolk (approved):

PR #391, opened 2021-09-08
Make union order deterministic

EMattfolk:
And remove a lot of code that was not needed.


FredTheDino (approved):
Nice! I too hate code.

sornas (approved):
What's better than code? Less code.

PR #392, opened 2021-09-08
Rework the syntax to use begin-end and more...

FredTheDino:
- Do-End blocks parsing
- Generate `do-end`
- fixup! Do-End blocks parsing
- Fix the easy cases
- Aggressive test renaming so I don't have to run the formatter
- Only accept do end-blocks
- Fix even more tests
- Fix the last tests


sornas:
Re warnings as a separate build: it is actually kind of nice. It makes it easy to see if it's just the warnings, or if there are faulty tests.

sornas:
It should at least not be hidden in an unrelated PR.

sylt-parser/src/statement.rs
```diff
@@ -549,7 +554,7 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
 
     // Newline, RightBrace and Else can end a statment.
     // If a statement does not end, we only report it as a missing newline.
-    let ctx = if matches!(ctx.token(), T::RightBrace | T::Else) {
+    let ctx = if matches!(ctx.token(), T::RightBrace | T::End | T::Else) {
```
EMattfolk, line 557: Can RightBrace end a statement anymore?

sornas (approved):

EMattfolk (approved):
Now we just need syntax highlighting

PR #395, opened 2021-09-09
Check nulls!

FredTheDino:
A very simple first step, there's more to do, like removing the int type from else-statements, and handling "and" in the condition etc...

### DIFF
--- a/sylt-compiler/src/typechecker.rs
+++ b/sylt-compiler/src/typechecker.rs
@@ -91,7 +91,7 @@ enum Name {
 
 #[derive(Debug, Clone)]
 enum Lookup {
-    Value(Type, VarKind),
+    Value { ty: Type, kind: VarKind, slot: Option<usize> },
     Namespace(usize),
 }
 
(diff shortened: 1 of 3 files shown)


sylt-compiler/src/typechecker.rs
```diff
@@ -923,7 +923,44 @@ impl<'c> TypeChecker<'c> {
...
+                            }
+                        }
+                        _ => Err(()),
+                    }
+                }
+
+                let check = null_check(self, condition);
+                if let Ok((ty, slot)) = check {
+                    let old_ty = self.stack[slot].ty.clone();
+                    self.stack[slot].ty = ty;
```
EMattfolk, line 956: Should'nt the null check create a new variable with the same name and a different type, sort of like shadowing, instead of mutating the stack? That would allow the variable to be changed inside a if statement (for example) without requiring the variable struct to be changed(?).

tests/typing/simple_if_null_check.sy
```diff
@@ -0,0 +1,22 @@
...
+        ret nil
+    end
+end
+
+start :: fn do
+    a := f(true)
+
+    if :int is a do
+        a = a + 1
+    end
```
EMattfolk, line 14: Does this work for blobs as well?

sylt-compiler/src/typechecker.rs
FredTheDino: This should simplify the implementation as well... I can't think of a counter argument, so I should try it!

tests/typing/simple_if_null_check.sy
FredTheDino: If you mean blob field accesses, no. If you mean union types with blobs, yes it should work.

sylt-compiler/src/typechecker.rs
FredTheDino: But this entire solution - of changing the lookup, is kinda limited and won't work for blob-fields. I think we as a whole want a different solution where we store extra information that we've deduced from the code, so I don't know how long this code will actually stay.

FredTheDino: Third times the charm - I don't know how useful it would be to not modify the variable, since this code would have to be changed to do a more `proper version` anyways. There are already a ton of problems with this. It doesn't handle:
 - globals
 - blob-fields
 - extenensions

So I've change my mind for the third time in 9 minutes. It's a good idea, but the code won't stay for long enough for the change to be worth it.

EMattfolk: I mean, this is not something that is very easy to do right. Not many languages have this feature except maybe TypeScript. It might be worth looking at what they do to solve this.

sornas:
todo: now that #399 is merged we can check if this change does what we think it does

FredTheDino:
Does not work, we need to reevaluate things when they stop working.

FredTheDino:
This is not complex enough to work properly - closing.

Issue #396, opened 2021-09-09
Really bad syntax error


FredTheDino:
```
start :: fn ->
    if true do
      print("a")
    else
        print("c")
    end
end
```

Complains with:
```
      Syntax Error on line 7
      Assignable expressions have to start with an identifier
   5 |         print("c")
   6 |     end
   7 | end
       ^^^
```

FredTheDino:
This error still persists.

PR #397, opened 2021-09-09
Don't start with end/else

FredTheDino:
Founds a bad syntax error (#396) I didn't manage to solve it, I think we might have to change how if-statements work. I'll have to think more about it tomorrow. But this makes the error, a little less awful.


sornas (commented):
Change looks good but it doesn't build.

EMattfolk (approved):

sornas (approved):

PR #399, opened 2021-09-10
Add faulty union op test

sornas:


sylt-compiler/src/typechecker.rs
```diff
@@ -1144,19 +1143,6 @@ mod op {
...
-            .find_map(|x| {
-                let x = f(x);
-                if x.is_nil() {
-                    None
-                } else {
-                    Some(x)
-                }
-            })
-            .unwrap_or(Type::Invalid)
-    }
```
EMattfolk, line 1158: This would still be valid when the un_op is `-` and the type is `int | float`. But I guess we would have to specify that somehow?

EMattfolk (approved):

sylt-compiler/src/typechecker.rs
sornas: That's a very good point. It feels like a different PR.

PR #402, opened 2021-09-15
Blobs rework

EMattfolk:
Instead of storing the blob values in a separate Vec we now store all the type information in the type itself.


sylt-common/src/flat_value.rs
```diff
@@ -13,8 +13,8 @@ pub type FlatValueID = usize;
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub enum FlatValue {
     Ty(Type),
-    Blob(usize),
-    Instance(usize, HashMap<String, FlatValueID>),
+    Blob(Type),
```
FredTheDino, line 16: Why can't this be the Type-type?

sylt-common/src/ty.rs
```diff
@@ -82,8 +82,8 @@ impl Display for Type {
                 }
                 write!(f, " -> {}", ret)
             }
-            Type::Blob(id) => write!(f, "Blob({})", id),
-            Type::Instance(id) => write!(f, "Instance({})", id),
+            Type::Blob(name, _) => write!(f, "{}", name),
+            Type::Instance(name, _) => write!(f, "{}", name),
```
FredTheDino, line 86: We want to differentiate these somehow, they need to be different.

FredTheDino (approved):
Top kek

PR #411, opened 2021-09-18
Simplify types

### LINKED ISSUE #385: Format `int | nil` as `int?`

FredTheDino:
Relates to #385 
```
start :: fn do
    a : int | void = 1
    a : float | void | int = 1
    b : void = nil
end
```

### DIFF
--- a/sylt-parser/src/expression.rs
+++ b/sylt-parser/src/expression.rs
@@ -4,7 +4,7 @@ use crate::statement::block;
 
 use super::*;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub enum ComparisonKind {
     Equals,
     NotEquals,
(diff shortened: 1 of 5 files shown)


EMattfolk (commented):
This program produces an invalid program. Not sure what is going on here, but it might be a good idea to fix it...
```
start :: fn do
    a :  void | void | int | float = 1
end
```

sornas:
Sounds like a useful test to have.

FredTheDino:
The failing test is because we have problems in our type-syntax.

```
fn -> fn -> int | int
```
Can be read in 2 different ways.
It's either:
 - A function (a) returning a function (b), (b) returns either an int or an in
 - A function (a) returning a function (b) (which returns an int) or an int
 
In "Haskell":
```
Unit -> (Unit -> Either Int Int)
Unit -> Either (Unit -> Int) Int
```

sornas (approved):

PR #412, opened 2021-09-18
Formatter tweaks

FredTheDino:
Changed some things that bugged me about the formatter:
 - Statements now write out their own whitespace, this fixes the trailing whitespace bug and the combining of multiple blank statements work better
 - Blobs and Instances write their constructor on one line, if it's only 1 thing and now share code
 - Add a space in one of @sornas comments :3

### DIFF
--- a/sylt/src/formatter.rs
+++ b/sylt/src/formatter.rs
@@ -59,11 +59,28 @@ fn write_blob_fields<W: Write>(
     indent: u32,
     fields: Vec<(String, Expression)>,
 ) -> fmt::Result {
-    for (field, expr) in fields {
-        write_indents(dest, indent)?;
-        write!(dest, "{}: ", field)?;
-        write_expression(dest, indent, expr)?;
-        write!(dest, ",\n")?;
+    write!(dest, " {{ ")?;
+    match fields.len() {
+        0 => {
+            write!(dest, " }}")?;
+        }
+        1 => {
+            let (field, expr) = fields[0].clone();
+            write!(dest, "{}: ", field)?;
+            write_expression(dest, indent, expr)?;
+            write!(dest, " }}")?;
+        }
+        _ => {
+            write!(dest, "\n")?;
+            for (field, expr) in fields {
+                write_indents(dest, indent)?;
+                write!(dest, "{}: ", field)?;
+                write_expression(dest, indent, expr)?;
+                write!(dest, ",\n")?;
+            }
+            write_indents(dest, indent - 1)?;
+            write!(dest, "}}")?;
+        }
     }
     Ok(())
 }
@@ -326,13 +343,7 @@ fn write_expression<W: Write>(dest: &mut W, indent: u32, expression: Expression)
         }
         ExpressionKind::Blob { blob, fields } => {
             write_assignable(dest, indent, blob)?;
-            write!(dest, " {{")?;
-            if !fields.is_empty() {
-                write!(dest, "\n")?;
-                write_blob_fields(dest, indent + 1, fields)?;
-                write_indents(dest, indent)?;
-            }
-            write!(dest, "}}")?;
+            write_blob_fields(dest, indent + 1, fields)?;
         }
         ExpressionKind::Tuple(exprs) => {
             let num_exprs = exprs.len();
--- a/sylt/src/formatter.rs
+++ b/sylt/src/formatter.rs
@@ -516,24 +516,16 @@ fn write_statement<W: Write>(dest: &mut W, indent: u32, statement: Statement) ->
 /// Replace consecutive empty statements with one empty statement with all comments of the previous statements.
 //TODO(gu): Rewrite the formatter to use moves instead of borrows. Then we wouldn't need to clone when passing
 //          into this function.
-fn merge_empty_statements(mut statements: Vec<Statement>) -> Vec<Statement> {
-    // Reverse since
-    // - we always want to remove and look at the first statement and
-    // - pop() is faster than remove(0).
-    statements.reverse();
-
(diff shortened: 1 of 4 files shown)


sylt/src/formatter.rs
```diff
@@ -54,16 +54,34 @@ fn write_parameters<W: Write>(
...
+            write!(dest, "\n")?;
+            for (field, t) in fields {
+                write_indents(dest, indent)?;
+                write!(dest, "{}: ", field)?;
+                sub_write(dest, indent, t)?;
+                write!(dest, ",\n")?;
+            }
+            write_indents(dest, indent - 1)?;
+            write!(dest, "}}")?;
+        }
```
EMattfolk, line 84: I think this writes trailing whitespace when there is two fields or more. Line 63 and 75.

sylt/src/formatter.rs
```diff
@@ -54,16 +54,34 @@ fn write_parameters<W: Write>(
     Ok(())
 }
 
-fn write_blob_fields<W: Write>(
+fn write_blob_fields<T: Clone, W: Write>(
```
sornas, line 57: I can't add a suggestion here since there's mixed additions and deletions, so consider the following diff:
```diff
diff --git a/sylt/src/formatter.rs b/sylt/src/formatter.rs
index 235cd22..afa390f 100644
--- a/sylt/src/formatter.rs
+++ b/sylt/src/formatter.rs
@@ -54,10 +54,10 @@ fn write_parameters<W: Write>(
     Ok(())
 }

-fn write_blob_fields<T: Clone, W: Write>(
+fn write_blob_fields<T, W: Write>(
     dest: &mut W,
     indent: u32,
-    fields: Vec<(String, T)>,
+    mut fields: Vec<(String, T)>,
     sub_write: fn(&mut W, u32, T) -> fmt::Result,
 ) -> fmt::Result {
     write!(dest, " {{ ")?;
@@ -66,7 +66,7 @@ fn write_blob_fields<T: Clone, W: Write>(
             write!(dest, " }}")?;
         }
         1 => {
-            let (field, expr) = fields[0].clone();
+            let (field, expr) = fields.remove(0);
             write!(dest, "{}: ", field)?;
             sub_write(dest, indent, expr)?;
             write!(dest, " }}")?;
```
Since we're using owned data we shouldn't need to clone anywhere.

sylt/src/formatter.rs
```diff
@@ -454,19 +480,20 @@ fn write_statement<W: Write>(dest: &mut W, indent: u32, statement: Statement) ->
             }
             write_expression(dest, indent, value)?;
         }
-        StatementKind::EmptyStatement => (),
+        StatementKind::EmptyStatement => {},
```
sornas, line 483: Either `(),` or `{}` but not both. Personally, I prefer `(),` since it's on one line.

sornas (commented):
I'd love to see in what comment I had forgotten a space too.

sylt/src/formatter.rs
```diff
@@ -509,33 +544,29 @@ fn write_statement<W: Write>(dest: &mut W, indent: u32, statement: Statement) ->
...
+            let is_empty = matches!(stmt.kind, EmptyStatement);
+            let no_comments = stmt.comments.is_empty();
+            if last_is_empty && is_empty && no_comments {
+                (ret, prev)
+            } else {
+                ret.push(stmt.clone());
+                (ret, Some(stmt))
+            }
+        }
+    ).0
```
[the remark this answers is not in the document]
sornas: I think the fold is hard to read, and also we don't need to clone here.

sylt/src/formatter.rs
```diff
@@ -509,33 +544,29 @@ fn write_statement<W: Write>(dest: &mut W, indent: u32, statement: Statement) ->
...
-            Some(StatementKind::EmptyStatement)
-        ) {
-            statement
-                .comments
-                .append(&mut statements.pop().unwrap().comments);
-        }
-        ret.push(statement);
-    }
-    ret
+// TODO(gu): Rewrite the formatter to use moves instead of borrows. Then we wouldn't need to clone when passing
```
FredTheDino, line 553: @sornas

sornas: I see. You'll find that all my todos are written this way.

sylt/src/formatter.rs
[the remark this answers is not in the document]
FredTheDino: I don't agree that the fold is hard to read. I could simplify some things, but look at it again. ;)

sylt/src/formatter.rs
FredTheDino: Fixed 'em for you ;)

sylt/src/formatter.rs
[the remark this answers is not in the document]
sornas: I think it's better, but I still think it's better to use `filter` when you want to filter instead of using `fold`.

sylt/src/formatter.rs
```diff
@@ -509,33 +543,25 @@ fn write_statement<W: Write>(dest: &mut W, indent: u32, statement: Statement) ->
             }
         }
     }
+    write!(dest, "\n")?;
 
     Ok(())
 }
 
 /// Replace consecutive empty statements with one empty statement with all comments of the previous statements.
```
sornas, line 551: I believe this doc comment is wrong now.

sylt/src/formatter.rs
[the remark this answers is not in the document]
FredTheDino: One way to look at it, is that I filter the list. Another way is to say that I `fold` away the duplicate statements.

If it weren't for the filter requiring side-effects, I'd be onboard. I just don't like side-effects in my filters if I can avoid id. And I can.

sornas (approved):

PR #414, opened 2021-09-18
grouping for types

FredTheDino:
- Format Void in unions as ?
- Add really good test
- Agressive test
- Make things PartialEq
- Simplify type
- Not today!
- Chicken out of the change and keep what can be kept
- TypeKind::Grouping
- Solve the parsing problems
- Fix some tests

### DIFF
--- a/sylt-parser/src/parser.rs
+++ b/sylt-parser/src/parser.rs
@@ -541,33 +541,33 @@ pub fn parse_type<'t>(ctx: Context<'t>) -> ParseResult<'t, Type> {
             let mut ctx = ctx.skip(1);
             let mut types = Vec::new();
             // Tuples may (and probably will) contain multiple types.
+            let mut is_tuple = matches!(ctx.token(), T::Comma | T::RightParen);
             loop {
+                // Any initial comma is skipped since we checked it before entering the loop.
+                ctx = ctx.skip_if(T::Comma);
                 match ctx.token() {
-                    // Done parsing this tuple.
-                    T::RightParen => {
-                        ctx = ctx.skip(1);
+                    // Done.
+                    T::EOF | T::RightParen => {
                         break;
                     }
 
-                    T::EOF => {
-                        raise_syntax_error!(ctx, "Didn't expect EOF in type definition");
-                    }
-
-                    // Parse a single contained type.
+                    // Another inner expression.
                     _ => {
-                        let (_ctx, param) = parse_type(ctx)?;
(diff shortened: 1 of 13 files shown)


sylt-parser/src/parser.rs
```diff
@@ -182,6 +182,8 @@ pub enum TypeKind {
     Dict(Box<Type>, Box<Type>),
     /// A generic type
     Generic(Identifier),
+    /// `(inner_type)` - usefull to correct the grouping
```
EMattfolk, line 185:
```suggestion
    /// `(inner_type)` - useful for correcting ambiguous types
```

sylt-parser/src/parser.rs
```diff
@@ -539,33 +541,33 @@ pub fn parse_type<'t>(ctx: Context<'t>) -> ParseResult<'t, Type> {
             let mut ctx = ctx.skip(1);
             let mut types = Vec::new();
             // Tuples may (and probably will) contain multiple types.
+            let mut is_tuple = matches!(ctx.token(), T::Comma | T::RightParen);
```
EMattfolk, line 544: It shouldn't be able to start with a comma right?
Perhaps skip it at the end of the loop.

sylt-parser/src/parser.rs
```diff
@@ -1166,8 +1170,8 @@ impl Display for Type {
             TypeKind::Tuple(tys) => {
                 write!(f, "(")?;
                 for (i, ty) in tys.iter().enumerate() {
-                    if i != 0 { write!(f, ", ")?; }
-                    write!(f, "{}", ty)?;
+                    if i != 0 { write!(f, " ")?; }
+                    write!(f, "{},", ty)?;
```
EMattfolk, line 1174: I think this causes trailing commas.

sylt-parser/src/parser.rs
FredTheDino: It now matches the tuple syntax for values. As far as I know, this is correct.

sylt-parser/src/parser.rs
FredTheDino: Yes, but it's acceptable in tuples, I made a formatting decision that you want trailing commas in tuples.

FredTheDino: It simplifies the implementation for the debug prints. I don't care enough about it to add a special case. If you feel strongly make an issue.

PR #415, opened 2021-09-18
Group types in prints to avoid ambiguities

FredTheDino:


sylt-std/src/sylt.rs
```diff
@@ -327,7 +327,7 @@ sylt_macro::extern_function!(
     "sylt_std::sylt",
     sign,
     ? "Returns the sign of the value",
-    -> "fn #X -> #X", // TODO(ed): Figure out how we can limit x...
+    -> "(fn int -> int) | (fn float -> float)",
```
EMattfolk, line 330: I don't think this is something that can be called for reasons... We should discuss this a bit more.

FredTheDino: Wat...
```
start :: fn do
    print(sign(1))
end
```
This runs. I'd love to hear why.

FredTheDino: I think I understand what you're referring to.

There are programs that don't run that we view as valid. Yes.
For example:
```
f : !(fn int) | (fn str) : fn a: int -> int do a + 1 end
f("abc")
```
But this problem isn't new here. I view the culprit as the "force-type" operator. These new calling rules give meaning to calling union-functions. One can construct a simpler program: 
```
f : !str | int : 1
f + "abc"
```
Which won't run but we consider valid. This is a deliberate design in the type-system. And to get something more powerfull, we need a typesystem with higher order types. These higher order types feel out of scope, but can be added, though I'd rather see a complete rehash of the typesystem if we decide to go down that route.

This change solves a problem, but it isn't perfect. For that, we need MORE POWER.

PR #439, opened 2021-09-30
lua flag

FredTheDino:
Flag for compiling to lua


EMattfolk (approved):
Looks ok. We could just pipe the output to lua directly and skip any intermediate files. I guess we could even run lua from within the rust program.

PR #448, opened 2021-10-03
remove -- and ++ as tokens

sornas:


EMattfolk (approved):
I think this is a nice change, though I am not sure Lua will be happy with it (it does not allow --). As long as negation is a Lua function instead of syntax it should be fine.

sornas:
Wouldn't inserting spaces be good enough?

cc @FredTheDino

FredTheDino:
```
Lua 5.4.3  Copyright (C) 1994-2021 Lua.org, PUC-Rio
> - - - - 1
1
> - - - - - 1
-1
>
```
Looks to be - I don't think this is a problem honestly.

FredTheDino (approved):

PR #456, opened 2021-10-03
Remove the stuff that lua cannot handle

FredTheDino:
Remove some language features - which I didn't feel worked - and was hard to implement in Lua. It's a banger!


EMattfolk (commented):
I am all for removing these, but what about type inference? How do we check that something is of a specific type now? Isn't it impossible to distinguish unions?

FredTheDino:
Yes - unions are impossible to distinguish - and thus useless as arguments to user defined functions.

Issue #457, opened 2021-10-04
Invesigate the use_import.sy test


FredTheDino:
When I made the Lua-compiler - I ran the type-checker before the compiler (this wasn't needed but it's the way we want it so I decided to keep it). This caused a test to fail - and since it failed consistently accross lua and bytecode I decided to disable it. The test should run though.

I'm pretty sure it's related to imports and the order things are defined - but I don't know much more than that.

PR #458, opened 2021-10-04
Last tests for the lua-compiler

FredTheDino:
Finished of the last test for the lua-compiler. I had to disable one - but that's due to other errors which we'll have to solve later.

See #457


sylt/src/lib.rs
```diff
@@ -191,13 +191,38 @@ mod lua {
...
+
+                println!("Expect error: {}", $any_runtime_errors);
+                if $any_runtime_errors {
+                    assert_errs!(res, []);
+                } else {
+                    assert_errs!(res, $errs);
+                    return;
+                }
+
+                let mut file = file.clone();
```
sornas, line 211: unnecessary clone, i think

sylt/src/lib.rs
```diff
@@ -191,13 +191,38 @@ mod lua {
...
+                    .arg(file)
+                    .stdin(Stdio::null())
+                    .stderr(Stdio::piped())
+                    .stdout(Stdio::null())
+                    .output()
+                    .expect(concat!("Failed to run ", $path));
+                if $any_runtime_errors {
+                    assert!(!output.status.success(), "Program ran to completion when it should fail");
+                } else {
+                    assert!(output.status.success(), "Failed when it should succeed\n:STDERR:\n{:?}\n", String::from_utf8(output.stderr).unwrap_or("Even I don't understand this stderr :(".to_string()));
```
sornas, line 224: very cool!

sylt-macro/src/lib.rs
```diff
@@ -229,9 +232,11 @@ fn parse_test_settings(contents: String) -> TestSettings {
             if line.starts_with("@") {
                 line = format!("Error::SyntaxError {{ span: Span {{ line: {}, ..}}, .. }}", &line[1..]);
             }
+            settings.any_runtime_errors |= line.contains("RuntimeError");
             errors.push(line);
         } else if line.starts_with("// flags: ") {
             for flag in line.split(" ").skip(2) {
+                // This can give false positives - but they're pretty unlikely
```
sornas, line 239: some info here would be nice

sornas (approved):

sylt-compiler/src/preamble.lua
```diff
@@ -1,7 +1,22 @@
...
 -- Built in types
+
+-- THE nil-table
+__NIL = setmetatable( { "nil" }, { __tostring = function() return "nil" end } )
+
+__INDEX = function(o, i)
+    local m = getmetatable(o)
+    if m._type == "tuple" or m._type == "list" then
+        local e = o[i + 1]
+        assert(e ~= nil, "Tuple/list index out of range \"" .. i .. "\"")
```
EMattfolk, line 11: Is this a problem if we have optional types in a list?

tests/typing/faulty_if_expressions.sy
```diff
@@ -1,8 +0,0 @@
-start :: fn do
-    1.0 + 1.0 if true else 2.0
-    (1.0 if < 0.0 else 2.0) + 1.0
-    1.0 + 1.0 if "potato" else 2.0
-    (1.0 if < "abc" else 2.0) + 1.0
-end
-// error: $Mismatch { expected: Type::Bool, .. }
-// error: $BinOp { lhs: Type::Float, .. }
```
EMattfolk, line 8: These are still valid tests right? We should keep them (with some modifications).

EMattfolk (approved):
Looks good. I would like to see the tests pass before merging though.

sylt-compiler/src/preamble.lua
FredTheDino: No - because this is after the type-system, we have another nil-value than Nil

tests/typing/faulty_if_expressions.sy
FredTheDino: No - these are not valid. They tested a syntax and semantic that has been removed - the short-if-statements.

PR #460, opened 2021-10-05
Write lua to Write

FredTheDino:
Skips the file steps and pipes lua code directly into lua or to a file.

The rust-docs mentioned something about stdout blocking stdin - but that might not be the case for my lua-compiler. Maybe we should add a test for it?

I'm pretty sure my installed lua-compiler `lua` parses the entire file and then emits errors and the like - and it always returns the first error and only the first error. So I won't have this problem, and I don't see it as a problem for just the test-suite since lua-jit shouldn't be used there.

IDK - but it works now.

Also - I updated the CLI a bit, you can now write to stdout and compile to a file and pipe it directly into lua. 

> More features with 10 fewer lines


sylt-compiler/src/lua.rs
```diff
@@ -31,16 +29,15 @@ macro_rules! write {
...
 
 impl<'t> LuaCompiler<'t> {
-    pub(crate) fn new(compiler: &'t mut Compiler, file: &Path) -> Self {
-        let file = File::create(&file).unwrap();
+    pub(crate) fn new(compiler: &'t mut Compiler, file: Box<dyn Write>) -> Self {
         Self {
             compiler,
             loops: Vec::new(),
-            file,
+            file: Box::new(file),
```
EMattfolk, line 40:
```suggestion
            file,
```
I tested and it does not need two boxes.

sylt/src/lib.rs
```diff
@@ -211,17 +271,29 @@ mod lua {
...
-                let output = Command::new("lua")
-                    .arg(file)
-                    .stdin(Stdio::null())
-                    .stderr(Stdio::piped())
-                    .stdout(Stdio::null())
-                    .output()
-                    .expect(concat!("Failed to run ", $path));
+                let output = child.wait_with_output().unwrap();
+                // NOTE(ed): Status is always 0 - atleast on my version of lua - so we check
+                // for stderr - which is kinda a bad idea.
```
EMattfolk, line 276: This should be investigated more. I get non-zero statuses in my terminal at least.

EMattfolk (commented):
I think this works for now, though we should think about how the CLI is structured. Perhaps having subcommands like git and cargo is a good idea?

There is a test that fails in the pipeline, it looks like there is some compilation error? Not sure why cargo test does not catch it... The failing test is `tests/blob/cluster.sy`.

.github/workflows/coverage.yml
```diff
@@ -10,12 +10,20 @@ env:
...
   build:
-    timeout-minutes: 2
+    timeout-minutes: 10
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v2
 
+      - name: Install Lua
+        run: sudo apt update && sudo apt install -y lua5.3
```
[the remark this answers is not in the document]
sornas: I believe `-y` is to automatically answer yes to all questions. Basically `--non-interactive`.

sylt/src/lib.rs
sornas: I, too, get a non-zero status.

```
$ lua not-lua.lua
lua: not-lua.lua:1: syntax error near ';'
$ echo $status
1
$ cat not-lua.lua
aflasjga;lsgj
```

sornas:
```
Lua 5.4.3  Copyright (C) 1994-2021 Lua.org, PUC-Rio
```

sornas:
> Perhaps having subcommands like git and cargo is a good idea?

I'm a big fan of dependencies, what can I say. https://lib.rs/crates/seahorse

sylt/src/lib.rs
FredTheDino: It's ONLY when you pipe the file - I might not have been clear about that:
```
$ echo "not-lua&&%¤&%¤%" | lua
```

FredTheDino:
> > Perhaps having subcommands like git and cargo is a good idea?
> 
> I'm a big fan of dependencies, what can I say. [lib.rs/crates/seahorse](https://lib.rs/crates/seahorse)

Maybe - it looks like Seahorse does everything we want, might be worth taking a look at

EMattfolk (approved):
It looks ok and seems to be working. The failing test still need to be resolved, but that can be done later.

PR #468, opened 2021-10-08
Add test coverage for sylt_std

th3tard1sparadox:
No "close enough"-comparison function added in this PR.


sornas:
`clear` at https://github.com/FredTheDino/sylt-lang/issues/477

FredTheDino:
#415 added tests for sign - feel free to overwrite them :3

EMattfolk:
> Alright so `clear` hasn't been merged yet. I think the best for now is to remove it (the test for clear) from here and add it over in #485 instead.

I would say update it here and hope for the best. But whatever works.

PR #501, opened 2021-10-17
documentation navbar

sornas:
Issues:

- [x] Navbar hides header after following anchor link: https://github.com/sylt-lang/sylt-lang/pull/501#issuecomment-945088557

Images:

image

image

Stays on top when scrolling:

image

Hover color:

image

Small screens:

image


sornas:
Issue: the navbar hides the header when following an anchor link.

Expected:

image


Current: 

image


This would be solved by moving the navbar into the table of contents like Erik suggested, but it won't stay at the top when using a small-screen device like a phone. (Although, how much do we want to support phones anyways?)

PR #502, opened 2021-10-17
install bs4 in docs workflow

sornas:
oops


sornas:
maybe we should try to build (without publishing) in PRs

PR #513, opened 2021-10-30
Hindley-Milner type system

FredTheDino:
A complete revamp of the type-system. Now with more than 1000% more UnionFind.


sornas (approved):
lgtm :+1:

PR #515, opened 2021-10-31
Union find Edvards midnight adventure

FredTheDino:


sylt-compiler/src/typechecker.rs
```diff
@@ -791,6 +1006,86 @@ impl TypeChecker {
...
+                // NOTE(ed): If we get here - it means we're checking the constraint, but the
+                // constraint shouldn't be added because we only ever check constants.
+                return err_type_error!(
+                    self,
+                    span,
+                    ctx,
+                    TypeError::Exotic,
+                    "Tuples can only be indexed by integer constants"
+                );
+            }
```
FredTheDino, line 1063: This is actually wrong - we need to make sure the integer is constants - we need a marker that we can track throughout the codebase, or we will allow invalid programs when passing tuples to functions! D:

sylt-parser/src/parser.rs
```diff
@@ -1183,7 +1183,7 @@ impl Display for Type {
                 write!(f, "{{{}:{}}}", k, v)?;
             }
             TypeKind::Generic(name) => {
-                write!(f, "#{}", name.name)?;
+                write!(f, "*{}", name)?;
```
FredTheDino, line 1186: Sneaky bugger...

PR #518, opened 2021-11-01
Implement proper namespace access

EMattfolk:


sylt-compiler/src/typechecker.rs
```diff
@@ -1464,13 +1496,28 @@ impl TypeChecker {
...
-                    namespace: *namespace,
-                },
-            )?;
+            if matches!(statement.kind, StatementKind::Use { .. }) {
+                self.outer_statement(
+                    statement,
+                    TypeCtx {
+                        namespace: *namespace,
+                    },
+                )?;
```
FredTheDino, line 1507: Maybe move in the Use code here - and remove it from outer-statement?
Then the `matches!` would disapear. Which makes me warm and fuzzy inside!

FredTheDino (changes requested):

sylt-compiler/src/typechecker.rs
```diff
@@ -791,6 +791,8 @@ impl TypeChecker {
                     panic!();
                 }
 
+                self.stack.truncate(ss);
```
FredTheDino, line 794: Good idea

PR #525, opened 2021-11-04
Remove union-types

FredTheDino:


sylt-parser/src/parser.rs
```diff
@@ -673,29 +671,6 @@ pub fn parse_type<'t>(ctx: Context<'t>) -> ParseResult<'t, Type> {
     // Wrap it in a syntax tree node.
     let ty = Type { span, kind };
```
EMattfolk, line 673:
```suggestion
```
This could be inlined further down.

sylt-std/src/sylt.rs
```diff
@@ -79,7 +79,7 @@ sylt_macro::extern_function!(
     "sylt_std::sylt",
     reduce,
     ? "Reduce the list to a single element, returns 'nil' if the input list is empty",
-    -> "fn [*ITEM], fn *ITEM, *ITEM -> *OUT -> *OUT | nil",
+    -> "fn [*ITEM], (fn *ITEM, *ITEM -> *OUT) -> *OUT", // The return is a lie!
```
EMattfolk, line 82: Yes, this is why reduce is not as nice as fold.

sylt-std/src/sylt.rs
```diff
@@ -183,10 +183,13 @@ sylt_macro::extern_function!(
     "sylt_std::sylt",
     len,
     ? "Gives the length of list",
-    -> "fn [*ITEM] | {*KEY: *VALUE} -> int",
+    -> "fn<C: Container> *C -> int",
```
EMattfolk, line 186: :open_mouth: This does look pretty sick.

sylt-std/src/sylt.rs
```diff
@@ -326,7 +329,7 @@ sylt_macro::extern_function!(
     "sylt_std::sylt",
     sign,
     ? "Returns the sign of the value",
-    -> "(fn int -> int) | (fn float -> float)",
+    -> "fn<N: Num> *N -> *N",
```
EMattfolk, line 332: Finally!

sylt-std/src/sylt.rs
```diff
@@ -456,7 +461,7 @@ sylt_macro::extern_function!(
     "sylt_std::sylt",
     dot,
     ? "Computes the scalar product",
-    -> "fn (float, float) -> float",
+    -> "fn (float, float), (float, float) -> float",
```
EMattfolk, line 464: This one could be a num as well. But no need to change it now.

EMattfolk (approved):

sornas (commented):
looks great

sylt-compiler/src/typechecker.rs
```diff
@@ -268,7 +268,16 @@ impl TypeChecker {
...
-                _ => return err_type_error!(self, span, ctx, todo_error!()),
+                _ => {
+                    return err_type_error!(
+                        self,
+                        ident.span,
+                        ctx,
+                        TypeError::Exotic,
+                        "Cannot find type '{}' - is it perhaps a type-variable?",
+                        ident.name
+                    )
```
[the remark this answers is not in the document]
FredTheDino: Yes, why i added it. If you write a type like: `fn C -> *`
If you forget a type variable or the type isn't defined. Should probably not the exotic.

sylt-parser/src/parser.rs
FredTheDino: Go for it

sylt-std/src/sylt.rs
FredTheDino: We'd need to change the function, the break up vec function returns floats.

PR #526, opened 2021-11-04
Union find edvard fixes todo

FredTheDino:


sylt-compiler/src/typechecker.rs
```diff
@@ -117,7 +117,7 @@ enum Constraint {
     Container,
     SameContainer(usize),
     Contains(usize),
-    IsContainedBy(usize),
+    Holds(usize),
```
EMattfolk, line 120: Now holds and contains seem like the same thing.

tests/hm_typing/in_expressions.sy
```diff
@@ -0,0 +1,9 @@
+f :: fn a: *, b: * -> *
+    a in b
+end
+
+start :: fn do
+    true <=> f' 1, [1, 2, 3]
+    true <=> f' 1, {1, 2, 3}
+    false <=> f' 0, {1, 2, 3}
+end
```
EMattfolk, line 9: What about dicts? I feel like it would not work as expected.

sylt-compiler/src/typechecker.rs
FredTheDino: If you want to change it back, I'm fine with the old one.

tests/hm_typing/in_expressions.sy
FredTheDino: We can change the constraint, now you would check for tuples in it, unless we double the constraints.

PR #528, opened 2021-11-05
Mitigate dependency checking errors

EMattfolk:
This is more of a hotfix. But it should work. Ideally the Names would be unique, but that means untangling code that may be removed in the future, so it is best to wait.

### DIFF
--- a/sylt-compiler/src/dependency.rs
+++ b/sylt-compiler/src/dependency.rs
@@ -25,13 +25,17 @@ impl Context<'_> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


FredTheDino (changes requested):
I would like a test that fails with the old and passes with the new. A so called regression test.

Also - what is the `tangled code`? What is the structure change that we'd need?

EMattfolk:
> I would like a test that fails with the old and passes with the new. A so called regression test.

There already is one. access_chain.

> Also - what is the `tangled code`? What is the structure change that we'd need?

Names are spread to the VM, dependency checker and compiler. They would need to be changed somehow.

PR #531, opened 2021-11-06
Blob rework

EMattfolk:
Fixed most of the things with blobs.


sylt-compiler/src/typechecker.rs
```diff
@@ -671,7 +657,12 @@ impl TypeChecker {
                         Some(Name::Global(var)) => Ok(var.ty),
                         None => match self.functions.get(&ident.name) {
                             Some(f) => Ok(*f),
-                            None => panic!("Cannot read variable: {:?}", ident.name),
+                            None => err_type_error!(
+                                self,
+                                span,
+                                ctx,
+                                TypeError::UnresolvedName(ident.name.clone())
+                            ),
```
FredTheDino, line 665: NIce

sylt-compiler/src/typechecker.rs
```diff
@@ -885,30 +876,62 @@ impl TypeChecker {
...
+                            kind: VarKind::Const,
+                            ty: given_blob,
+                        });
+                    }
+                    let expr_ty = self.expression(expr, ctx)?;
+                    self.unify(span, ctx, expr_ty, given_fields[key])?;
+                    if is_function {
+                        self.stack.pop();
+                    }
+                }
```
FredTheDino, line 933: I wonder - sometimes late at night - if the better code truncates the stack to a certain size at the beginning of each loop.

sylt-compiler/src/typechecker.rs
```diff
@@ -1210,25 +1233,36 @@ impl TypeChecker {
...
+                            self,
+                            span,
+                            ctx,
+                            TypeError::Mismatch {
+                                got: self.bake_type(a),
+                                expected: self.bake_type(b)
+                            },
+                            "Blobs have different number of fields"
+                        );
+                    }
```
FredTheDino, line 1247: I dislike this error message. We can do better by pointing to the actual fields that are wrong. This will not really help anyone. :(
In a perfect world we'd search through the fields and look for a close word, and then add that as a guess. It will save me not noticing typos. :3

Copy the bellow for-loop and reverse it. It has the added benefit of making the code symetric.

FredTheDino (changes requested):

sylt-compiler/src/typechecker.rs
EMattfolk: No.

sylt-compiler/src/typechecker.rs
EMattfolk: We do that in the expression. Perhaps it is a good idea to do it here too. Ideally we would only show the differences. I don't think we thought about that when we first implemented it. There might be a nice solution.

sylt-compiler/src/typechecker.rs
```diff
@@ -1210,25 +1233,36 @@ impl TypeChecker {
                 }
 
                 (Type::Blob(a_blob, a_fields), Type::Blob(b_blob, b_fields)) => {
-                    let mut c_fields = BTreeMap::new();
-                    for (a_name, a_ty) in a_fields.iter() {
-                        let b_ty = match b_fields.get(a_name) {
+                    if a_fields.len() != b_fields.len() {
```
FredTheDino, line 1234:
```suggestion
                    // TODO(ed): This should give information about the violating fields.
                    if a_fields.len() != b_fields.len() {
```

PR #544, opened 2021-11-07
Even more function declaration helpers

FredTheDino:
- Implicit unknown if not specified
- Drive-by: _ctx -> ctx_
- Special case for empty functions
- Fix argument syntax with implicit return
- Solve implicit return if using ->


EMattfolk (commented):
I would like to see some more tests. Maybe something with `fn -> () end` or similar.

FredTheDino:
> I would like to see some more tests. Maybe something with `fn -> () end` or similar.

This test is really messy to add - it's not a parser test, since it's a valid function signature (the `()` is parsed as the return type) - but it won't typecheck since it doesn't return `()` but rather `nil`.

This PR still expects a type after the arrow, followed by statements - so the parsing hasn't changed. For this change to be made we had to do #545 

TLDR: I added the test but it does nothing

EMattfolk (approved):

PR #545, opened 2021-11-07
TypeAssignable

FredTheDino:
- Tokens for types and type-identifiers
- Parser for type-assignables
- Add in type-assignable
- Simplify parsing
- Remove stale todo


EMattfolk (approved):
Looks good, though I don't really see what it will be used for.

PR #547, opened 2021-11-07
unit instead of nil

### LINKED ISSUE #476: Remove nil?
The unit type `()` could possibly be used instead.

FredTheDino:
- Change the syntax so () -> nil
- Special cases in the test fixed

This doesn't change the internal representation - because the `void`/`()` type
is very helpfull when typechecking and reasoning about the code. This is basically
a reskin - which kinda makes me question why we want this. I don't know if I want it still, since it doesn't solve a single problem - only makes the syntax harder to parse.

Closes #476

### DIFF
--- a/sylt-parser/src/expression.rs
+++ b/sylt-parser/src/expression.rs
@@ -256,7 +256,6 @@ fn value<'t>(ctx: Context<'t>) -> Result<(Context<'t>, Expression), (Context<'t>
         T::Float(f) => Float(f),
         T::Int(i) => Int(i),
         T::Bool(b) => Bool(b),
-        T::Nil => Nil,
         T::String(s) => Str(s),
         t => {
             raise_syntax_error!(ctx, "Cannot parse value, '{:?}' is not a valid value", t);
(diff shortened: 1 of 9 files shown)


sylt-parser/src/expression.rs
```diff
@@ -484,7 +483,7 @@ fn infix<'t>(ctx: Context<'t>, lhs: &Expression) -> ParseResult<'t, Expression>
 /// one-sized tuple containing `1`.
 ///
 /// `()` as well as `(,)` are parsed as zero-sized tuples.
```
sornas, line 485: doc is wrong

sornas (commented):
code looks good but I have no strong preference for either `nil` or `()`.

sornas (approved):

EMattfolk:
> Did we gain a single thing from this?

Not really in the state we are in right now. But, by removing () as a type (meaning void has the only value ()) we can make types and values completely disjoint, which is a (very) good thing.
It would remove the ambiguity from expressions like `fn -> () end`. Maybe we could even do more fancy syntax changes...

drathier:
If you're going to change it, could you please semantically separate `void` and `unit`? Mathematically, void has no values and is a type that cannot be constructed. Unit has one value and can only be constructed in one way, so it carries no information.

EMattfolk:
> If you're going to change it, could you please semantically separate `void` and `unit`? Mathematically, void has no values and is a type that cannot be constructed. Unit has one value and can only be constructed in one way, so it carries no information.

Good point. We haven't really thought about what the bottom type is since we rewrote the type system. It used to be the `Invalid` type, but I don't think it is used anymore. Setting `void` to be the bottom type is probably a good idea.

FredTheDino:
A bottom type might be interesting. It could make functions that cannot return, which would be helpful to reason about flow-control.

I am not sure the complexity pays for itself though... I'm leaning a lot towards closing this PR tough - further discussions can be had in the issue

PR #550, opened 2021-11-08
Docs update

### LINKED ISSUE #543: update documentation to match new typesystem
- [ ] ducktyping
- [ ] deduce

more?

EMattfolk:
Closes #543

### DIFF
--- a/docs/guide.adoc
+++ b/docs/guide.adoc
@@ -44,7 +44,8 @@ NUMBERS :: []  // a constant declaration
 Constant variables (apart from functions) should be written in
 `SCREAMING_SNAKE_CASE`.
 
-The type of the variable is inferred but can be supplied.
+The type of the variable is inferred but can be supplied if you want to be
+explicit.
 
 === Functions
 
@@ -63,9 +64,9 @@ In this example we see two functions. First, we define a new function and store
 it in the constant variable `greet`. We also use the print-function in the
 function body and print a string to the terminal.
 
-The function block starts with `do` and ends with `end`. The `do` can be
-replaced with a `->` if the function returns something. See the following
-examples.
+The function block starts with `do` or `->` and ends with `end`. `do` indicates
+no return value while `->` means the function returns something. See the
+following examples.
 
 .Function that takes a single parameter.
 [source, sylt]
@@ -303,7 +304,7 @@ end
 
 `new_player` creates a `Player` with a built in update-function that
 only mutates the created `Player` blob. The `Player` blob is baked
-into the update-function as the constant `self`.
+into the update-function as `self`.
 
 There are a lot of fun things you can do with closures, but this
 language feature can easily be abused to create unreadable code.
@@ -460,7 +461,7 @@ l := [1, 2, 3]
 l -> for_each' print
 
 l -> for_each' fn a: int do
-    // ...
+    // something
 end
 ----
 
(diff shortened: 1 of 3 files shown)


docs/guide.adoc
```diff
@@ -44,7 +44,7 @@ NUMBERS :: []  // a constant declaration
 Constant variables (apart from functions) should be written in
 `SCREAMING_SNAKE_CASE`.
 
-The type of the variable is inferred but can be supplied or even forced.
+The type of the variable is inferred but can be supplied.
```
FredTheDino, line 47:
```suggestion
The type of the variable is inferred but can be supplied if you want to be explicit.
```

docs/guide.adoc
```diff
@@ -63,8 +63,9 @@ In this example we see two functions. First, we define a new function and store
 it in the constant variable `greet`. We also use the print-function in the
 function body and print a string to the terminal.
 
-The function block starts with `do` and ends with `end`. The `do` can optionally
-be omitted if the function has a return type. See the following examples.
+The function block starts with `do` and ends with `end`. The `do` can be
+replaced with a `->` if the function returns something. See the following
+examples.
```
FredTheDino, line 68:
```suggestion
The function block starts with `do` or `->`, and ends with `end`. `do` indicates no return value.
An `->` means the function returns something. See the following
examples.
```
It read a bit wierd - like arrows were special cases.

docs/guide.adoc
```diff
@@ -197,7 +198,7 @@ iterator functions.
 .Map and implicit `ret`
 [source, sylt]
 ----
-squares :: [1, 2, 3] -> map' fn a: int -> int a * a end
+squares :: [1, 2, 3] -> map' fn a -> a * a end
```
FredTheDino, line 201: Yum

docs/guide.adoc
```diff
@@ -302,7 +303,7 @@ end
 
 `new_player` creates a `Player` with a built in update-function that
 only mutates the created `Player` blob. The `Player` blob is baked
-into the update-function and cannot be changed after creation.
+into the update-function as the constant `self`.
```
FredTheDino, line 306:
```suggestion
into the update-function as the psuedo-variable `self`.
```
I feel like claiming `self` to be constant is a bit confusing, since you can change the fields (which you can do with constants, but I'd say it's not the main point)

docs/guide.adoc
```diff
@@ -456,10 +457,10 @@ mutable (e.g. another list) it can be mutated.
 ----
 l := [1, 2, 3]
 
-l -> for_each' fn a: int do print(a) end
+l -> for_each' print
 
 l -> for_each' fn a: int do
-    // many statements
+    // ...
```
FredTheDino, line 463:
```suggestion
    // many statements
```
Felt this was clearer - idk

docs/guide.adoc
```diff
@@ -497,7 +498,7 @@ points := [1, 2, 3]
 
 sum := 6
 
-points_str := points -> map' fn p: int -> str do
```
FredTheDino, line 500: You do know this is still allowed?

docs/guide.adoc
```diff
@@ -745,51 +716,50 @@ Spider :: blob {
...
+//though every element is treated as if being of type `Creature`.
+//
+//To call the `eat_bug` function anyway, forced types may be used.
+//
+//.Force general blob into specific blob
+//[source, sylt]
+//----
+//spider: !Spider = creatures[0]
+//spider.eat_bug() // Ok, as long as creatures[0] is a Spider
+//----
```
FredTheDino, line 762:
```suggestion
```
Ducktyping is dead as far as I'm concerned. There's no way to add it into the type-system as it is.

FredTheDino (approved):
Some minor nits - but looks good to me!

Maybe add a new section about type-inferance, and that it's totally awesome so you never have to write a type ever again?

docs/guide.adoc
sornas:
```
The function block starts with `do` or `->` and ends with `end`. `do` indicates
no return value while `->` means the function returns something. See the
following examples.
```

docs/guide.adoc
sornas: pseudo-variable is on the other hand a bit vague. I'd remove it all together.

```
The `Player` blob is baked
into the update-function as `self`.
```

The example makes it clear how it can be used.

docs/guide.adoc
```diff
@@ -456,10 +457,10 @@ mutable (e.g. another list) it can be mutated.
 ----
 l := [1, 2, 3]
 
-l -> for_each' fn a: int do print(a) end
+l -> for_each' print
```
sornas, line 460: pog

docs/guide.adoc
EMattfolk: Yes, but I don't see the point of being inconsistent. I'd say we remove that syntax, but it's not hurting anyone.

FredTheDino: Removing it is gonna be tricky - but I like the optionality of it. It might also be helpfull if we want to allow functions definitions over multiple lines. I can also see it being used as a stylistic approach when inlining with a return type:
```
add_one :: fn a -> int do a + 1 end
```
This reads clearer than:
```
add_one :: fn a -> int a + 1 end
```
If you want to explicitly write the return type.

docs/guide.adoc
```diff
@@ -63,9 +64,9 @@ In this example we see two functions. First, we define a new function and store
 it in the constant variable `greet`. We also use the print-function in the
 function body and print a string to the terminal.
 
-The function block starts with `do` and ends with `end`. The `do` can be
-replaced with a `->` if the function returns something. See the following
-examples.
+The function block starts with `do` or `->` and ends with `end`. `do` indicates
+no return value while `->` means the function returns something. See the
+following examples.
```
sornas, line 69: `do` doesn't require no return value. Although I can't seem to think of a good rewording.

sornas (approved):
last comment still stands but it's not enough to block imo

docs/guide.adoc
FredTheDino: I think the whole sentence has to be restructured

PR #553, opened 2021-11-08
Function copying

### LINKED ISSUE #549: Unification for functions behaves unexpectedly
```
start :: fn do
    [1, 2, 3] -> for_each' print
    print(1.0)
end

      A 'int' cannot be a 'float'
      Types don't match
   1 | start :: fn do
   2 |     [1, 2, 3] -> for_each' print
   3 |     print(1.0)
                ^
```
I hope this is not a serious problem...

EMattfolk:
Closes #549


sylt-compiler/src/typechecker.rs
```diff
@@ -635,12 +632,7 @@ impl TypeChecker {
 
                 let expression_ty = self.expression(value, ctx)?;
                 let defined_ty = self.resolve_type(span, ctx, &ty)?;
-                let expression_ty = if matches!(self.find_type(defined_ty), Type::Unknown) {
-                    // TODO(ed): Not sure this is needed
-                    self.copy(expression_ty)
-                } else {
-                    expression_ty
-                };
```
FredTheDino, line 643: me <- :1st_place_medal: me

FredTheDino (commented):
That last test is sneaky -  we'll have to add a "calling constraint" which resolves to the callable type afterwards - it's probably gonna be a bit messy.

EMattfolk:
I have been thinking about this. I think it may be very complicated. Even more so than what the sneaky tests shows. I have still no clue how higher order functions would work. Perhaps we just live with it?

```
f :: fn a: fn * -> void ->
    b : [*] : [a]
    b[0]' 1
    b[0]' "hello"
    a' "hello"
end

start :: fn do
    f' print
end
```
This works fine. Maybe we can detect when things like this happen and throw a helpful error?

FredTheDino:
I have an idea that might work - I'll test it later today if I get the time.

The idea is to add constraints for the arguments. Since information cannot flow backwards from constraints, and there is not need to add more type information to the function arguments. This will hurt deducing of return-types, which might be a problem, but I think it's our best shot.

EMattfolk:
What do we think about this? I would say we skip fixing this for now, keep the issue and fix it properly later. I don't think it has the highest priority right now.

FredTheDino (approved):
I think we should do a more permanent fix - or add in the error, but I don't really like half-fixes. It's gonna take time to fix the things we know are wrong now and to change the context to working with it. Especially since the fix feels so trivially small.

FredTheDino (approved):
Lgtm. We have a test for recursive functions right? Good that you added the function special case. Open an issue about this aswell

EMattfolk:
> Lgtm. We have a test for recursive functions right? Good that you added the function special case. Open an issue about this aswell

Yes, and I have already opened another issue about this.

PR #556, opened 2021-11-09
Reformat and merge tests to lessen clutter and compile times

FredTheDino:


EMattfolk (approved):
Looks good, though we should probably merge tests with coverage.

sornas (approved):

PR #557, opened 2021-11-09
Better spans

FredTheDino:
Added more information to the spans - so now we don't have to send in the ctx to write errors. Preventing death by a 1000 papercuts


FredTheDino:
Fused into #558

PR #558, opened 2021-11-09
Error and span refactor

FredTheDino:


sornas:
Failing tests :o

sylt-compiler/src/bytecode.rs
```diff
@@ -708,10 +701,10 @@ impl<'t> BytecodeCompiler<'t> {
 
         // TODO(ed): Real ugly hack until we can run the typechecker before the compiler.
```
EMattfolk, line 702: :eye:

sylt-compiler/src/typechecker.rs
```diff
@@ -1251,11 +1265,11 @@ impl TypeChecker {
 
         self.types[b].parent = Some(a);
         self.types[a].size += self.types[b].size;
-        self.types[a].constraints = self.types[a]
-            .constraints
-            .union(&self.types[b].constraints)
-            .cloned()
-            .collect();
+
+        // TODO(ed): Which span should we keep? The one closest to the top? Should we combine them?
```
EMattfolk, line 1269: Does it matter? It is the same constraint.

sylt-compiler/src/typechecker.rs
```diff
@@ -1400,34 +1410,45 @@ impl TypeChecker {
...
+                        Constraint::Sub(x) => Constraint::Sub(self.inner_copy(*x, seen)),
+                        Constraint::Mul(x) => Constraint::Mul(self.inner_copy(*x, seen)),
+                        Constraint::Div(x) => Constraint::Div(self.inner_copy(*x, seen)),
+                        Constraint::Equ(x) => Constraint::Equ(self.inner_copy(*x, seen)),
+                        Constraint::Cmp(x) => Constraint::Cmp(self.inner_copy(*x, seen)),
+                        Constraint::CmpEqu(x) => Constraint::CmpEqu(self.inner_copy(*x, seen)),
+                        Constraint::Neg => Constraint::Neg,
+                        Constraint::Indexes(x) => Constraint::Indexes(self.inner_copy(*x, seen)),
+                        Constraint::IndexedBy(x) => {
+                            Constraint::IndexedBy(self.inner_copy(*x, seen))
```
EMattfolk, line 1426: We should have some test with unknown dicts and lists.

sylt-compiler/src/typechecker.rs
```diff
@@ -1543,15 +1569,15 @@ impl TypeChecker {
...
                 TypeError::Assignability,
                 "Cannot assign to expressions"
             ),
         }
     }
 
-    fn add_constraint(&mut self, a: usize, constraint: Constraint) {
-        self.find_node_mut(a).constraints.insert(constraint);
+    fn add_constraint(&mut self, a: usize, span: Span, constraint: Constraint) {
+        // TODO(ed): Don't reinsert stuff?
```
EMattfolk, line 1579: :shrug:

sylt-compiler/src/typechecker.rs
```diff
@@ -14,30 +14,55 @@ use std::collections::{BTreeMap, BTreeSet};
...
+}
+
+impl<T> Help for TypeResult<T> {
+    fn help(mut self, typechecker: &TypeChecker, span: Span, message: String) -> Self {
+        match &mut self {
+            Ok(_) => {}
+            Err(errs) => match &mut errs.last_mut() {
+                Some(Error::TypeError { helpers, .. }) => {
+                    helpers.push(Helper {
+                        file: typechecker.namespace_to_file[&span.file_id].clone(),
```
EMattfolk, line 28: There must be some way of removing this...

EMattfolk (approved):
LG:tm:, though I have this feeling we can do even better.

sylt-compiler/src/typechecker.rs
FredTheDino: If we could have a global life-time, we could remove this. (But we'd need to mutate global state which is unsafe and will make Gustav very very sad)

FredTheDino:
> LGtm, though I have this feeling we can do even better.

I'm hoping to start merging spans aswell (why I added the multiple lines). This way we can point to a range of code and maybe make the error messages even better.

I'm assuming you're refering to clutter - I don't see a way to make this cleaner, but I'm not omnipotent (yet).

sylt-compiler/src/typechecker.rs
FredTheDino: Maybe it's a good idea to proratize spans that are in the same file? I feel like this can matter, and it feels like this can stay as a todo?

sylt-compiler/src/typechecker.rs
FredTheDino: Constraints that are added multiple times overwrite the span - should we keep the oldest span? Or is there some better heuristic for sources of constraints? They're all equivalent but there might be a better way to do all of this? Or maybe the newest one is the most relevant?

But we now don't overwrite things.

sylt-compiler/src/typechecker.rs
FredTheDino: Feel free to write those tests - these should not be a part of this PR though. This was just a formatting change.

PR #563, opened 2021-11-13
Typechecker minor fixes

sornas:


sylt-compiler/src/typechecker.rs
```diff
@@ -631,7 +625,10 @@ impl TypeChecker {
             | StatementKind::Blob { .. }
             | StatementKind::IsCheck { .. }
             | StatementKind::ExternalDefinition { .. } => {
-                unreachable!("Illegal inner statement! Parser should have caught this.")
+                unreachable!(
+                    "Illegal inner statement at {:?}! Parser should have caught this.",
+                    span
```
FredTheDino, line 630: Isnt this debug print is really bad? Maybe say which line and rsnge it has instead?

sylt-compiler/src/typechecker.rs
```diff
@@ -1499,24 +1486,17 @@ impl TypeChecker {
...
-                        )
-                        .help(
-                            self,
-                            var.span,
-                            "Originally defined here".into(),
-                        )
-                    }
-                } else {
-                    match self.globals.get(&(ctx.namespace, ident.name.clone())) {
+                match self.stack.iter().rfind(|v| v.ident.name == ident.name) {
```
FredTheDino, line 1489: Nice!

FredTheDino (approved):
Lgtm, nothing controversial! Maybe we should add a better print for spans?

EMattfolk (approved):

sylt-compiler/src/typechecker.rs
sornas: kinda bad, yes, but i don't know how to get the filename from just the span.

FredTheDino: You can check the type_error macro - the span has a `file_id` that indexes `namespace_to_filename` (I think these names are right, I might be wrong.)

FredTheDino:
This code is getting outdated - I'd recomend to merge it ASAP <3

sylt-compiler/src/typechecker.rs
```diff
@@ -655,7 +654,12 @@ impl TypeChecker {
             | StatementKind::StatementExpression { .. }
             | StatementKind::Unreachable
             | StatementKind::EmptyStatement => {
-                panic!("Illegal outer statement! Parser should have caught this")
+                panic!(
```
EMattfolk, line 657:
```suggestion
                unreachable!(
```
This is supposed to unreachable right?

EMattfolk (approved):

sylt-compiler/src/typechecker.rs
sornas: yes :+1:

PR #565, opened 2021-11-14
From imports

EMattfolk:


sylt-parser/src/parser.rs
```diff
@@ -888,7 +888,7 @@ fn module(
             Ok((ctx, statement)) => {
                 use StatementKind::*;
                 // Yank `use`s and add it to the used-files list.
```
FredTheDino, line 890: This comment is kinda half out of date

sylt-parser/src/statement.rs
```diff
@@ -302,6 +311,46 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
+                // in the folder.
+                parent.join(if path == "/" {
+                    format!("exports.sy")
+                } else if path_ident.name.ends_with("/") {
+                    format!("{}/exports.sy", name)
+                } else {
+                    format!("{}.sy", name)
+                })
+            };
+            let ctx = expect!(ctx, T::Use, "Expected 'use' after path");
```
FredTheDino, line 339: I like to mention what we expected to parse - so a clearer syntax error might be:
```suggestion
            let ctx = expect!(ctx, T::Use, "Expected 'use' after path when parsing from-import");
```

sylt-parser/src/statement.rs
```diff
@@ -302,6 +311,46 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
+            let mut names = Vec::new();
+            let ctx = match ctx.tokens_lookahead::<1>() {
```
FredTheDino, line 341: `ctx.token()` - is that you?

sylt-parser/src/statement.rs
```diff
@@ -302,6 +311,46 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
+                        name: name.clone(),
+                        span: ctx.span(),
+                    }));
+                    ctx.skip(1)
+                }
+                _ => raise_syntax_error!(ctx, "Expected identifier"),
+            };
+            (ctx, From { path: path_ident, names, file })
+        }
+
```
FredTheDino, line 353: Add a parse test for good measure! :D

FredTheDino (commented):
Looks like a good start - just the semantics left! <3

sylt-parser/src/statement.rs
```diff
@@ -302,6 +311,46 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
+                // Importing a folder is the same as importing exports.sy
+                // in the folder.
+                parent.join(if path == "/" {
+                    format!("exports.sy")
+                } else if path_ident.name.ends_with("/") {
+                    format!("{}/exports.sy", name)
+                } else {
+                    format!("{}.sy", name)
+                })
+            };
```
FredTheDino, line 338: This looks very similar to the importing done in `Use` - maybe fuse these together into the path-function?

FredTheDino (commented):
Sry... Looked at it again...

sylt-compiler/src/typechecker.rs
```diff
@@ -618,6 +619,29 @@ impl TypeChecker {
                     .insert((ctx.namespace, ident.name.clone()), Name::Namespace(other));
             }
 
+            StatementKind::From { imports, file, .. } => {
+                // TODO(ed): This shouldn't be nessecary - the namespace should be set up
+                // Correctly already.
```
FredTheDino, line 624:
```suggestion
                // correctly already.
```

sylt-parser/src/statement.rs
```diff
@@ -33,6 +33,15 @@ pub enum StatementKind {
         file: PathBuf,
     },
 
+    /// "Imports" variables from another file.
+    ///
+    /// `from <file> use <var1>
+    From {
+        path: Identifier,
+        imports: Vec<(Identifier, Option<Identifier>)>,
```
FredTheDino, line 41: `Option<Identifier>` looks a bit cryptic... Maybe make a type alias for this? Or use a struct, since they're often easier to read. Might not be super clear that the option is for renaming.

sylt-parser/src/statement.rs
FredTheDino: Poke

sylt-parser/src/statement.rs
```diff
@@ -302,6 +311,84 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
+                                T::Identifier(name) => {
+                                    Some(Identifier { name: name.clone(), span: ctx.span() })
+                                }
+                                _ => raise_syntax_error!(ctx, "Expected identifier after 'as'"),
+                            };
+                            ctx = ctx.skip(1);
+                            _alias
+                        } else {
+                            None
+                        };
```
FredTheDino, line 368: Maybe this should be broken out - so it can be used in the other import aswell? It's gonna be a lot easier to keep this consistent if it's the same parsing function.

sylt-parser/src/statement.rs
```diff
@@ -640,13 +728,19 @@ mod test {
     test!(outer_statement, outer_statement_use_rename: "use a as b\n" => _);
     test!(outer_statement, outer_statement_use_subdir: "use a/b/c/d/e\n" => _);
     test!(outer_statement, outer_statement_use_subdir_rename: "use a/b as c\n" => _);
+    test!(outer_statement, outer_statement_from: "from a/b use c\n" => _);
+    test!(outer_statement, outer_statement_from_many: "from b use c,d,e,f,g,h\n" => _);
+    test!(outer_statement, outer_statement_from_paren: "from / use (c\n,d\n)\n" => _);
+    test!(outer_statement, outer_statement_from_paren_one: "from / use (c)\n" => _);
```
FredTheDino, line 734: <3

sylt-parser/src/statement.rs
```diff
@@ -302,6 +311,84 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
+                    T::RightParen | T::Newline => break,
+                    T::Identifier(name) => {
+                        let ident = Identifier { name: name.clone(), span: ctx.span() };
+                        ctx = ctx.skip(1);
+                        let alias = if matches!(ctx.token(), T::As) {
+                            ctx = ctx.skip(1);
+                            let _alias = match ctx.token() {
+                                T::Identifier(name) => {
+                                    Some(Identifier { name: name.clone(), span: ctx.span() })
+                                }
```
FredTheDino, line 361: We should add a check for the capitalization of the renamed name and the original name to be the same - since capitals implies a type. I.e, this should be illegal: `from a import B as b` and vice versa. Might be out of scope for this PR though.

FredTheDino (approved):
The tests and logic look good to me - but I'd love to see some refactoring and merging of the other import code. The parser is slowly getting out of hand and we should get on that ASAP.

sylt-parser/src/statement.rs
sornas: it could also be an enum.

```rust
pub enum Import {
    Normal(Identifier),
    Renamed(Identifier, Identifier),
    // or
    Renamed {
        old: Identifier
        new: Identifier,
    }
}

File: sylt-compiler/src/dependency.rs [hunk 11]
```diff
@@ -225,6 +229,33 @@ fn statement_dependencies(ctx: &mut Context, statement: &Statement) -> BTreeSet<
...
+                            kind: AssignableKind::Read(name.clone()),
+                        },
+                    )
+                })
+                .flatten()
+                .collect();
+            ctx.namespace = old_ns;
+            deps
+        }
+
```
FredTheDino, line 258: We can split up the tree before this - so we can solve more dependencies.

sylt/src/formatter.rs
```diff
@@ -544,6 +544,32 @@ fn write_statement<W: Write>(dest: &mut W, indent: u32, statement: Statement) ->
...
+                    write_identifier(dest, alias.clone())
+                } else {
+                    Ok(())
+                }
+            }
+            write_indents(dest, indent)?;
+            write!(dest, "from ")?;
+            write_identifier(dest, path)?;
+            write!(dest, " use ")?;
+            if imports.len() == 1 {
```
FredTheDino, line 565: Sort this alphabetically?

FredTheDino (approved):

PR #567, opened 2021-11-15
From imports multiple

FredTheDino:
Added multiple imports - we might still need to support `as` in the imports, this should also be simple to add in. :D


sylt-compiler/src/dependency.rs
```diff
@@ -373,6 +383,19 @@ pub(crate) fn initialization_order<'a>(
         for statement in module.statements.iter() {
             use StatementKind::*;
             match &statement.kind {
+                From { imports, .. } => {
+                    let mut ctx = Context { compiler, namespace, variables: Vec::new() };
+                    imports.iter().for_each(|Identifier { name, .. }| {
+                        to_order.insert(
+                            (name.clone(), namespace),
+                            (
+                                statement_dependencies(&mut ctx, statement),
```
EMattfolk, line 392: This is not really true since it means every variable in the from import must be initialized.
We could probably replace this with a read assignable in the other namespace...

sylt-parser/src/statement.rs
```diff
@@ -336,15 +336,41 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
+            loop {
+                match ctx.token() {
+                    T::RightParen | T::Newline => break,
+                    T::Identifier(name) => {
+                        imports.push(Identifier { name: name.clone(), span: ctx.span() });
+                        ctx = ctx.skip(1);
+                        if matches!(ctx.token(), T::Comma | T::RightParen) {
+                            syntax_error!(ctx, "Expected ',' after imports");
+                        }
+                        ctx = ctx.skip_if(T::Comma);
```
EMattfolk, line 359: This seems very wrong, but it works?

EMattfolk: Took a closer look and it seems to just not do anything.

FredTheDino: It skips if it's a comma! :D Which is exactly what the code says.

sylt-compiler/src/dependency.rs
FredTheDino: It shouldn't make much difference

FredTheDino (commented):
Feel free to merge this in and fix the stuff yourself - I'll work more on enums, which feels more pressing.

sylt-parser/src/statement.rs
EMattfolk: I mean the error. Anyway, it should be a quick fix.

sylt-compiler/src/compiler.rs
```diff
@@ -460,6 +461,41 @@ impl Compiler {
...
+                                    "A global variable with the name '{}' already exists",
+                                    ident_name
+                                );
+                            }
+                        }
+                    }
+                }
+                _ => unreachable!(),
+            }
+        }
```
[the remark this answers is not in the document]
EMattfolk: It is pretty ugly. At least it ended up in the ugliest function. I think this is just one of those parts that need the complexity.

PR #590, opened 2021-11-24
Parser fixes

EMattfolk:
Fixed a small thing

### DIFF
--- a/sylt-parser/src/expression.rs
+++ b/sylt-parser/src/expression.rs
@@ -806,6 +806,7 @@ mod test {
     fail!(expression, list_funky: "[1 2]" => _);
     fail!(expression, set_funky: "{1 2}" => _);
     fail!(expression, dict_funky: "{1: 2 3: 4}" => _);
+    fail!(expression, tuple_funky: "(1 2 3 4}" => _);
 }
 
 impl PrettyPrint for Expression {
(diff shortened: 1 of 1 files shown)


sylt-parser/src/expression.rs
```diff
@@ -798,6 +803,9 @@ mod test {
     test!(expression, fn_invalid_empty_tuple: "fn -> () end" => _);
 
     fail!(expression, fn_self_arg: "fn self: int do 1 end" => _);
+    fail!(expression, list_funky: "[1 2]" => _);
+    fail!(expression, set_funky: "{1 2}" => _);
+    fail!(expression, dict_funky: "{1: 2 3: 4}" => _);
```
FredTheDino, line 808:
```suggestion
    fail!(expression, dict_funky: "{1: 2 3: 4}" => _);
    fail!(expression, tuple_funky: "(1 2 3 4}" => _);
```
Add tuples for completeness.

FredTheDino (approved):
But maybe we should question this? We could allow it since it isn't strictly required for the syntax, worth to ponder in my opinion.

(We should add tests for newlines in lists and sets aswell...)

EMattfolk:
I mean, I'm not strictly opposed to the syntax, but it still think it is not a good idea.

PR #593, opened 2021-11-24
More error information

FredTheDino:
Tries to be descriptive and transparent about `where` things don't match. We now get errors like this:
```
typecheck error: a.sy:8
      A 'str' cannot be a 'int'
      Types don't match
   6 |
   7 | start :: fn do
   8 |     A {
           ^
help:
      While checking dictionary value
      While checking field .q
```

It would be better to point to the actual field - but this cannot be done since types don't have positions in the source code. :(

(We could potentially match the expressions against the blob-fields in the blob-expression, but that might be a bit too... extreem?)

### DIFF
--- a/sylt-compiler/src/typechecker.rs
+++ b/sylt-compiler/src/typechecker.rs
@@ -1503,19 +1503,24 @@ impl TypeChecker {
                 (Type::Str, Type::Str) => {}
 
                 (Type::List(a), Type::List(b)) => {
-                    self.unify(span, ctx, a, b)?;
+                    self.unify(span, ctx, a, b)
+                        .help_no_span("While checking list".into())?;
                 }
                 (Type::Set(a), Type::Set(b)) => {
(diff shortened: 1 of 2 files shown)


EMattfolk (approved):

sylt-compiler/src/typechecker.rs
```diff
@@ -24,7 +25,23 @@ impl<T> Help for TypeResult<T> {
             Ok(_) => {}
             Err(errs) => match &mut errs.last_mut() {
                 Some(Error::TypeError { helpers, .. }) => {
-                    helpers.push(Helper { file: typechecker.span_file(&span), span, message });
+                    helpers.push(Helper {
+                        at: Some((typechecker.span_file(&span), span)),
+                        message,
+                    });
+                }
+                _ => panic!("Cannot help on this error"),
```
sornas, line 33: :thinking:

sornas (approved):

PR #594, opened 2021-11-25
Recursive types

FredTheDino:


sylt-compiler/src/typechecker.rs
```diff
@@ -777,10 +782,10 @@ impl TypeChecker {
...
                 };
                 let ty = match self.globals.get(&(ctx.namespace, enum_name.name.clone())) {
-                    Some(Name::Enum(ty)) => {
-                        // NOTE(ed): This allows enumerations to be generic, which is always fun!
-                        let ty = ty.clone();
-                        let ty = self.push_type(ty);
+                    Some(Name::Type(ty)) => {
+                        // NOTE(ed): This allows generics, which is always fun!
+                        let ty = *ty;
+                        let ty = self.copy(ty);
```
EMattfolk, line 788:
```suggestion
                        let ty = self.copy(*ty);
```
It is just a usize right?

EMattfolk (commented):
Looks ok, though I am not sure what has happened to the git history. Maybe fix that before merging?

sylt-compiler/src/typechecker.rs
FredTheDino: I tried this - it doesn't work since it's a borrow from self and Rust gets sad and sucks.

sornas (approved):

Issue #595, opened 2021-11-25
Typechecker bug? from import use blob


Quaqqer:
Importing Blobs with `from <import> use <blob>` causes an error, I assume that this is a bug. I provided some example test cases below. `from_use_blob.sy` does not work while `use_blob.sy` works. The error that `from_use_blob.sy` outputs is the following:

```
typecheck error: from_use_blob.sy:4
      Cannot resolve name 'Bob'
      Expected a blob or an enum
   2 | 
   3 | start :: fn do
   4 |     bob := Bob {
                  ^^^
```

https://github.com/sylt-lang/sylt-lang/compare/main...Quaqqer:blob-the-builder

EMattfolk:
Nice find! I will check it out. Hopefully it is just some small thing we missed.

EMattfolk:
It does indeed look like something we missed. Maybe you want to fix it?
Look at line 662 it the typechecker. There the blobs, enums and namespaces are ignored. The fix would be initializing them similar to the other match branches in `outer_statement`. You could cram in some more tests as well.

PR #600, opened 2021-11-25
Fix `from <import> use <blob|enum>`

Quaqqer:
Resolves issue #595.

I believe this is working. I crammed in a few more tests, although I am unsure if they are very good.

### DIFF
--- a/sylt-compiler/src/typechecker.rs
+++ b/sylt-compiler/src/typechecker.rs
@@ -676,7 +676,7 @@ impl TypeChecker {
                 let other = self.file_to_namespace[file];
                 for (ident, alias) in imports.iter() {
                     let name = self.globals[&(other, ident.name.clone())].clone();
-                    let (namespace, identifier) = (ctx.namespace, alias.as_ref().unwrap_or(ident));
+                    let ident_name = &alias.as_ref().unwrap_or(ident).name;
 
                     match name {
                         Name::Global(other_var) => {
@@ -688,15 +688,22 @@ impl TypeChecker {
                             };
                             self.unify(span, ctx, var.ty, other_var.ty)?;
                             self.globals
-                                .insert((namespace, identifier.name.clone()), Name::Global(var));
+                                .insert((ctx.namespace, ident_name.clone()), Name::Global(var));
                         }
 
                         Name::Type(_) => {
                             self.globals
-                                .insert((namespace, identifier.name.clone()), name.clone());
+                                .insert((ctx.namespace, ident_name.clone()), name.clone());
                         }
 
-                        Name::Namespace(_) => continue,
+                        Name::Namespace(_) => {
+                            return err_type_error!(
+                                self,
+                                span,
+                                TypeError::Exotic,
+                                "From import of namespaces is not implemented"
+                            );
+                        }
                     }
                 }
             }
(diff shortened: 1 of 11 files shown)


sornas:
image
nice!. try running `cargo fmt` again.

FredTheDino (approved):
Looks sicko to me!

sylt-compiler/src/typechecker.rs
```diff
@@ -657,21 +657,29 @@ impl TypeChecker {
...
+                            self.globals
+                                .insert((namespace, identifier.name.clone()), Name::Global(var));
+                        }
+
+                        Name::Blob(_) | Name::Enum(_) => {
+                            self.globals
+                                .insert((namespace, identifier.name.clone()), name.clone());
+                        }
+
+                        Name::Namespace(_) => ()
```
EMattfolk:
```suggestion
                        Name::Namespace(_) => todo!("From import of namespaces is not implemented")
```
Or similar. We should probably decide if it should even be possible.

EMattfolk (approved):
Looks good!

sylt-compiler/src/typechecker.rs
```diff
@@ -657,21 +657,29 @@ impl TypeChecker {
...
-                        kind: VarKind::Const,
-                        span,
-                    };
-                    self.unify(span, ctx, var.ty, other_var.ty)?;
-                    self.globals.insert(
-                        (ctx.namespace, alias.as_ref().unwrap_or(ident).name.clone()),
-                        Name::Global(var),
-                    );
+                    let name = self.globals[&(other, ident.name.clone())].clone();
+                    let (namespace, identifier) = (ctx.namespace, alias.as_ref().unwrap_or(ident));
```
FredTheDino, line 661: This is a bit funky - maybe break it up into two lines?

sylt-compiler/src/typechecker.rs
```diff
@@ -657,21 +657,29 @@ impl TypeChecker {
...
+                        Name::Global(other_var) => {
+                            let var = Variable {
+                                ident: alias.as_ref().unwrap_or(ident).clone(),
+                                ty: self.push_type(Type::Unknown),
+                                kind: VarKind::Const,
+                                span,
+                            };
+                            self.unify(span, ctx, var.ty, other_var.ty)?;
+                            self.globals
+                                .insert((namespace, identifier.name.clone()), Name::Global(var));
```
FredTheDino, line 673: Looks like you only use `identifier.name`, maybe read the on line 661 aswell?

FredTheDino (commented):
*me being tired and leaving comments after approving*

sylt-compiler/src/typechecker.rs
FredTheDino: This should probable give an actual error - since it is an error in userspace. Don't panic - throw an error, `Exotic` is maybe the kind you want.

FredTheDino:
Might also be worth to rebase...

Quaqqer:
I rebased into 3 commits from 6, removing the formatting ones for instance, do you want me to squash them all to one?

EMattfolk:
It is fine to keep it like that. We will probably merge this when the conflicts are fixed.

PR #614, opened 2021-11-28
Exaustive case patterns

### LINKED ISSUE #613: Exaustive case-statements ;D

FredTheDino:
- Add it.
- Fix the tests
Closes #613


sylt-compiler/src/bytecode.rs
```diff
@@ -782,7 +786,11 @@ fn all_paths_return(statement: &Statement) -> bool {
         StatementKind::If { pass, fail, .. } => all_paths_return(pass) && all_paths_return(fail),
 
         StatementKind::Case { branches, fall_through, .. } => {
-            branches.iter().all(|b| all_paths_return(&b.body)) && all_paths_return(fall_through)
+            branches.iter().all(|b| all_paths_return(&b.body))
+                && fall_through
+                    .as_ref()
+                    .map(|f| all_paths_return(f))
+                    .unwrap_or(true)
```
EMattfolk, line 793: Should this move to the typechecker? A comment above would indicate so.

sylt-compiler/src/typechecker.rs
```diff
@@ -1475,6 +1483,50 @@ impl TypeChecker {
...
+
+                    Type::Enum(enum_name, enum_vars) => {
+                        let mut missing = Vec::new();
+                        for var in vars.iter() {
+                            if !enum_vars.contains_key(var) {
+                                missing.push(var.clone());
+                            }
+                        }
+                        if !missing.is_empty() {
+                            err_type_error!(
```
EMattfolk, line 1498: Return the errors and flatten out the if-statements?

sylt-compiler/src/typechecker.rs
```diff
@@ -1475,6 +1483,50 @@ impl TypeChecker {
...
+                Constraint::TotalEnum(vars) => match self.find_type(a) {
+                    Type::Unknown => Ok(()),
+
+                    Type::Enum(enum_name, enum_vars) => {
+                        let mut missing = Vec::new();
+                        for var in vars.iter() {
+                            if !enum_vars.contains_key(var) {
+                                missing.push(var.clone());
+                            }
+                        }
```
EMattfolk, line 1496: This could probably be replaced with a filter and collect.

sylt-compiler/src/typechecker.rs
```diff
@@ -1475,6 +1483,50 @@ impl TypeChecker {
...
+                                span,
+                                TypeError::MissingVariants(enum_name.clone(), missing)
+                            )
+                        } else {
+                            let mut extra = Vec::new();
+                            for (var, _) in enum_vars.iter() {
+                                if !vars.contains(var) {
+                                    extra.push(var.clone());
+                                }
+                            }
```
EMattfolk, line 1509: This one too.

EMattfolk (approved):
Looks good. Just some minor things.

It is kind of sad that we always get worse coverage when adding better errors. I should look into if there is some way of testing the messages themselves.

FredTheDino:
> Looks good. Just some minor things.
> 
> It is kind of sad that we always get worse coverage when adding better errors. I should look into if there is some way of testing the messages themselves.

How worth is it? I honestly think the coverage is fundementally a flawed metric, but we've already talked about that. 

We fire each of the error paths, I honestly think it's just a problem with how coverage for rust is calculated. If you want to put time on it, feel free, but i think bugs are a better time investment right now.

sylt-compiler/src/bytecode.rs
FredTheDino: Don't touch the bytecode...

sylt-compiler/src/typechecker.rs
FredTheDino: We cannot do this, since we add help to all errors in unify. We cannot use return.

EMattfolk (approved):

sornas (approved):

PR #615, opened 2021-11-29
Changes from Fall Game Jam 2021

EMattfolk:
This should be everything, but I may have missed some things.


sylt-compiler/src/typechecker.rs
```diff
@@ -574,6 +574,7 @@ impl TypeChecker {
                         );
                         self.stack.push(var);
                     }
+                    self.check_constraints(span, ctx, to_match)?;
```
FredTheDino, line 577: We should add tests for this!!!

sylt-compiler/src/preamble.lua
```diff
@@ -282,13 +282,13 @@ end
 
 __BLOB_META = { _type = "blob" }
 __BLOB_META.__eq = function(a, b)
-    for k, _ in pairs(a) do
-        if not b[k] then
+    for k, v in pairs(a) do
+        if b[k] == nil or v ~= b[k] then
```
FredTheDino, line 286: We should test this!!! And we shouldn't need to test for null... Righ?

FredTheDino (commented):
Looks good, but we need to test more!

sylt-compiler/src/typechecker.rs
EMattfolk: I can't remember what it fixed :nog:

FredTheDino (approved):
LGTM!

PR #616, opened 2021-11-30
Fix prime calling for good

### LINKED ISSUE #608: Calling with `'` is strange at times
```
f :: fn -> 1 end
start :: fn do
    {f'}
    [f']
end
```

FredTheDino:
Closes #608

Does the obvious thing, and makes it work, and stuff!
(Might be too permissive but I can't think of any right now)


sornas (approved):

EMattfolk (approved):
I can't think of a counter-example either, but it seems to work well.

Issue #605, opened 2021-11-27
Typechecker does not catch tuples of different length?


EMattfolk:
```
start :: fn do
    ((1,) in {(1,2)}) <=> true
end
```
It might also be the `in` operator. Or something else entirely.

PR #618, opened 2021-11-30
Fix tuple lengths

FredTheDino:
Closes pr #605

### DIFF
--- a/sylt-common/src/error.rs
+++ b/sylt-common/src/error.rs
@@ -145,6 +145,11 @@ pub enum TypeError {
         length: usize,
     },
 
+    TupleLengthMismatch {
+        got: usize,
+        length: usize,
+    },
+
     UnresolvedName(String),
 
     WrongConstraintArity {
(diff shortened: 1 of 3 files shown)


sylt-common/src/error.rs
```diff
@@ -417,6 +422,14 @@ impl fmt::Display for TypeError {
                 write!(f, "A tuple of length {} has no element {}", length, got)
             }
 
+            TypeError::TupleLengthMismatch { length, got } => {
+                write!(
+                    f,
+                    "Tuple lengths don't match '{}'!='{}'",
```
sornas, line 428:
```suggestion
                    "Tuple lengths don't match, expected {} but got {}",
```

sornas (approved):

tests/bugs/tuple_length_mismatch_605.sy
```diff
@@ -0,0 +1,4 @@
+start :: fn do
+    ((1,) in {(1,2)}) <=> true
+end
+
```
EMattfolk, line 4:
```suggestion
// error: $TupleLengthMismatch {..}
```
Or something.

EMattfolk (approved):

sylt-common/src/error.rs
FredTheDino: Expected and got doesn't really make sense, since they're just two tuples that don't share lengths. The naming is actually wrong.

PR #619, opened 2021-11-30
Some blob comparison stuff

Quaqqer:
- Compare blobs in vm, might work incorrectly when subcomparison returns Value::Nil.
- Add blob comparison test
- Remove comparison with nil from lua preamble, might be wrong of me to do this in this commit since it was a comment in #615.

### DIFF
--- a/tests/blob/comparison.sy
+++ b/tests/blob/comparison.sy
@@ -1,0 +1,54 @@
+Blob :: blob {
+    a: bool,
+    b: int,
+    c: float,
+}
+
+BlobBlob :: blob {
+    _blob: Blob,
+}
+
+start :: fn do
+    b1 := Blob {
+        a: true,
+        b: 1,
+        c: 0.,
+    }
+
+    b1_eq := Blob {
+        a: true,
+        b: 1,
+        c: 0.,
+    }
+
+    b2 := Blob {
+        a: false,
+        b: 420,
+        c: .0,
+    }
+
+    b1 <=> b1_eq
+    b1 != b2 <=> true
+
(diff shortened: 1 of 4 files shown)


tests/blob/comparison.sy
```diff
@@ -0,0 +1,54 @@
...
+        _blob: b1_eq,
+    }
+
+    bb1 <=> bb1_eq2
+
+    bb2 := BlobBlob {
+        _blob: b2,
+    }
+
+    bb1 != bb2 <=> true
```
FredTheDino, line 53: It would be nice to see a negative test as well - so a case where they don't match - just to make sure we haven't made a tautology...

FredTheDino (approved):
Other than one test missing, this LGTM!

FredTheDino:
Also, maybe rename `comparison` to `equality`. I associate `comparisons` with `<` and `>`.

tests/blob/comparison.sy
Quaqqer: I'm not sure that I follow, isn't the assertion on line 53 a negative test since it checks for inequality between `bb1` and `bb2`? Or do you want me to add `bb1 == bb2 <=> false`? I'm not the most well read on testing.

EMattfolk: I think it looks good. There are plenty of negative tests, maybe Edvard was just tired?

EMattfolk (approved):

tests/blob/comparison.sy
FredTheDino: I was tired....

PR #623, opened 2021-12-01
Remove VM once and for all

FredTheDino:
- Remove the machine
- Nuke flat_value, network and lingon
- Add std/std.sy
- Refactor out the stdlib
- Add different kinds of sources
- Trust that we see that the arrows are wrong
- Make the tests run again - there's more that needs cutting though

Does 2 big things at once - it removes the vm and most things related to it. Somethings I skipped to remove.

I also added a different kind of "source" for reading in the source code - this source can then be the internal `std.sy` files and similar in the future. The commits are pretty good, so if you want to only see the source code changes view the last 3 commits.


sornas (approved):

sylt-parser/src/parser.rs
```diff
@@ -966,8 +966,9 @@ fn module(
             Ok((ctx, statement)) => {
                 use StatementKind::*;
                 // Get the used files from 'use' and 'from' statements.
-                if let Use { file, .. } | FromUse { file, .. } = &statement.kind {
-                    use_files.push(file.clone());
+                match &statement.kind {
+                    Use { file, .. } | FromUse { file, .. } => use_files.push(file.clone()),
+                    _ => {}
```
EMattfolk, line 971: This just seems like more code.

sylt-parser/src/statement.rs
```diff
@@ -211,28 +211,37 @@ pub fn path<'t>(ctx: Context<'t>) -> ParseResult<'t, Identifier> {
...
     let (ctx, path_ident) = path(ctx)?;
     let path = &path_ident.name;
     let name = path
         .trim_start_matches("/")
         .trim_end_matches("/")
         .to_string();
     let file = {
-        let parent = if path.starts_with("/") {
-            ctx.root
+        if let Some(name) = library_name(&name) {
```
EMattfolk, line 222: Does this prioritize libraries? I think the files should be prioritized whenever possible

sylt/src/lib.rs
```diff
@@ -136,15 +88,12 @@ pub struct Args {
...
     pub dump_tree: bool,
 
-    #[options(short = "l", long = "lua", help = "Run using lua")]
-    pub lua_run: bool,
-
     #[options(
         short = "c",
         long = "compile",
-        help = "Compile to a lua file - % for stdout"
+        help = "Compile to a lua file - for stdout"
```
EMattfolk, line 94:
```suggestion
        help = "Compile to a lua file, '-' for stdout"
```

EMattfolk (approved):

sylt-parser/src/statement.rs
FredTheDino: This can be changed, I'd prefer a warning

PR #634, opened 2021-12-08
Fix writes before compiling

### LINKED ISSUE #602: Do not write files if compilation fails
It causes several other problems. We should also think about updating the compiler interface.

EMattfolk:
Fixes #602


FredTheDino (approved):
LGTM

PR #641, opened 2021-12-11
Trim some unused things from the compiler

EMattfolk:


FredTheDino (approved):
Lgtm - maybe we can refactor even more here. :0

EMattfolk:
> Lgtm - maybe we can refactor even more here. :0

I just did :wink:, though I still think there is more to be done...

FredTheDino (approved):
LGTM

sornas (approved):

PR #645, opened 2021-12-28
Disallow instancing of void

FredTheDino:
We now don't allow this program:
```
start :: fn do
    f :: fn do end
    q :: f'
end
```


sylt-compiler/src/typechecker.rs
```diff
@@ -1529,6 +1535,17 @@ impl TypeChecker {
                         vars
                     ),
                 },
+
+                Constraint::Variable => match self.find_type(a) {
+                    Type::Void => err_type_error!(
+                        self,
+                        span,
+                        TypeError::Exotic,
```
EMattfolk, line 1543: As it is right now this error is quite common.

sylt-compiler/src/typechecker.rs
```diff
@@ -1529,6 +1535,17 @@ impl TypeChecker {
...
+                        "The `void` has no values and cannot be put in a variable"
```
EMattfolk, line 1544:
```suggestion
                        "The 'void' type has no values and cannot be put in a variable"
```
I don't remember any backticks in error messages...

sylt-parser/src/statement.rs
```diff
@@ -466,10 +466,9 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
         // `ret <expression>`
```
EMattfolk, line 466:
```suggestion
        // `ret [<expression>]`
```

EMattfolk (approved):

sylt-compiler/src/typechecker.rs
FredTheDino: Habits die hard - but it should probably be " actually...

sylt-parser/src/statement.rs
FredTheDino: The syntax has changed, the expression is optional now - so that's what the comment is signifying. Is it unclear?

sylt-compiler/src/typechecker.rs
FredTheDino: But not common in the code ;)

sylt-compiler/src/typechecker.rs
FredTheDino: Actually - this is concistent with the other error messages, see lines 658, 647. Not gonna change.

sylt-common/src/error.rs
```diff
@@ -421,6 +423,14 @@ impl fmt::Display for TypeError {
                 }
                 Ok(())
             }
+
+            TypeError::ExpectVoid(ty) => {
+                write!(
+                    f,
+                    "Only void types can unify with void - '{}' is not void",
```
sornas, line 430: maybe a bit technical

sornas (approved):

PR #646, opened 2021-12-28
Better tests!

FredTheDino:
Less macros - more speed - better test output.

Unfortunately we cannot run specific tests (since that's a cargo feature as far as I know). Feel free to come with opinions about the output. (Did some googling - and we can probably pass arguments somehow to the test-program :D)


sylt/src/test.rs
```diff
@@ -0,0 +1,258 @@
...
+    use std::io::Write;
+    set_current_dir("../").unwrap();
+    let tests = find_and_parse_tests(Path::new("tests/"));
+    let mut failed = Vec::new();
+
+    writeln!(
+        std::io::stdout().lock(),
+        "= RUNNING {} TESTS =",
+        tests.len()
+    )
```
sornas, line 219:
```suggestion
    writeln!(
        std::io::stdout().lock(),
        "= RUNNING {} PROGRAM TESTS =",
        tests.len()
    )
```
or something along those lines. we have other tests in addition to these

tests/conflict_markers_multiple.sy
```diff
@@ -12,5 +12,5 @@ start :: fn do
     print extern_test(5.0)
 >>>>>>> 2
 end
-// error: Error::GitConflictError { span: Span { line_start: 3 , .. }, .. }
-// error: Error::GitConflictError { span: Span { line_start: 9 , .. }, .. }
+// error: Git conflict marker found at line 3
+// error: Git conflict marker found at line 9
```
sornas, line 16: :eyes:

tests/dependencies/global_with_same_name.sy
```diff
@@ -1,3 +1,3 @@
 a :: 0
 a :: 0
-// error: Error::CompileError { .. }
+// error: A global variable with the name 'a' already exists
```
sornas, line 3: this is great

sornas (approved):

EMattfolk (approved):

PR #649, opened 2022-01-11
optimization a la edvard

FredTheDino:
We now use WAY less variables - the compilers time-complexity
hasn't changed but the memory usage has increased a lot for larger programs.

The steps are as follows:
 - Count the usages of each variable
 - Apply the following rule based on the count:
     - 0: remove
     - 1: inline (save the expression as a string and write it later)
     - 2+: use a variable
 - When writing code, we try to build the expressions as big as possible

There are some limitations with this.
We don't get optimal dead-code elimination, since a usage is a usage no matter
what. Consider this code:
```sylt
a :: 1
b :: a
c :: b
```
This we can only simplify to:
```sylt
a :: 1
b :: a

```
Since all other variables are used once in the original.
Though we could substitute it with:
```sylt



```

### DIFF
--- a/sylt-compiler/src/intermediate.rs
+++ b/sylt-compiler/src/intermediate.rs
@@ -73,6 +73,7 @@ pub enum IR {
 
     Label(Label),
     Goto(Label),
+    Copy(Var, Var),
     Define(Var),
     Assign(Var, Var),
     Return(Var),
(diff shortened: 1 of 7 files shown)


sylt-compiler/src/intermediate.rs
```diff
@@ -7,13 +7,12 @@ use sylt_parser::{
...
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq)]
 pub struct Var(pub usize);
 
-impl Display for Var {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let Var(n) = self;
-        write!(f, "V{}", n)
+impl Var {
+    pub fn format(self) -> String {
+        format!("V{}", self.0)
```
EMattfolk, line 15: This should probably still be the Display trait. `to_string` could be used instead since it gets automatically implemented.

EMattfolk (approved):
This works with my AoC solutions as well which is a good indicator :tada:

Maybe add some tests that would fail before? Like a large list or expression.

sornas (approved):
cool stuff

PR #653, opened 2022-01-12
Std improvements

### LINKED ISSUE #650: `std` improvements
- [x] Determine what should be where
- [x] Separate current std into logical modules
- [x] Rewrite what is possible in sylt
- [x] Default import heavily used parts of `std`
- [ ] Files with same name as library should be an error

EMattfolk:
Closes #650


sylt-parser/src/parser.rs
```diff
@@ -1054,6 +1054,7 @@ where
     // Files we want to parse but haven't yet.
     let mut to_visit = Vec::new();
     let root = path.parent().unwrap();
+
```
FredTheDino, line 1057: :eye:

sylt-parser/src/parser.rs
```diff
@@ -1086,13 +1087,37 @@ where
         let tokens = string_to_tokens(file_id, &source);
         // Parse the module.
         let (mut next, result) = module(&include, file_id, &root, &tokens);
+
```
FredTheDino, line 1090: :eye:

sylt-parser/src/parser.rs
```diff
@@ -1086,13 +1087,37 @@ where
...
     }
 
+    if bundle_std {
+        modules = modules
+            .into_iter()
+            .enumerate()
+            .map(|(i, (file, mut modul))| {
+                match file {
+                    FileOrLib::File(_) => {
+                        let tokens = string_to_tokens(i, include_str!("../../std/basics.sy"));
```
FredTheDino, line 1105: This can be moved out of the loop.

sylt-parser/src/parser.rs
```diff
@@ -1086,13 +1087,37 @@ where
...
+                        let (_, std) = module(&FileOrLib::Lib("basics"), i, &root, &tokens);
+                        let std = std.expect("Failed to parse default imports");
```
FredTheDino, line 1107: You can probably unpack the error to get better error messages

sylt-parser/src/parser.rs
```diff
@@ -1086,13 +1087,37 @@ where
...
+                        modul.statements = [std.statements.clone(), modul.statements]
+                            .iter()
+                            .cloned()
+                            .flatten()
+                            .collect();
```
FredTheDino, line 1112: I think this can be simplified. `modul` (which should probably be module) could probably append the `std.statements` to it, maybe even using: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append

sylt/src/formatter.rs
```diff
@@ -638,6 +638,6 @@ fn format_module(module: Module) -> Result<String, fmt::Error> {
 }
 
 pub fn format(file: &Path) -> Result<String, Vec<Error>> {
-    let mut tree = sylt_parser::tree(file, crate::read_file)?;
+    let mut tree = sylt_parser::tree(file, crate::read_file, false)?;
```
FredTheDino, line 641: I feel like this is a bit arbitrary? Maybe we should mark the imports in the syntax-tree as being default imported? If this flag should exist, it should be hooked up to the argument-parsing aswell - but maybe you've just not had time for that.

std/basics.sy
```diff
@@ -0,0 +1,42 @@
+from common use (
```
FredTheDino, line 1: Rename the automatically imported one to `prelude` - basics and common are to similar in meaning. I don't easily know which one is the auto-imported one. 

Maybe add a comment about only use-statements being allowed in the prelude?

sylt-compiler/src/preamble.lua
```diff
@@ -438,47 +440,15 @@ function sign(x)
         return 0
     end
 end
-function clamp(x, lo, hi)
-    return math.min(hi, math.max(x, lo))
-end
-min = math.min
-max = math.max
```
FredTheDino, line 445: Nice to move away from lua! :D

std/math.sy
```diff
@@ -0,0 +1,61 @@
...
+
+normalize :: fn v: (float, float) -> (float, float)
+    mag :: magnitude(v)
+    (v[0] / mag, v[1] / mag)
+end
+
+reflect: fn (float, float), (float, float) -> (float, float) : external
+
+dot :: fn v1: (float, float), v2: (float, float) -> float
+    ret v1[0] * v2[0] + v1[1] * v2[1]
```
FredTheDino, line 60:
```suggestion
    v1[0] * v2[0] + v1[1] * v2[1]
```
Don't need this ret

std/common.sy
```diff
@@ -0,0 +1,22 @@
...
+    ret x
+end
+
+args: fn -> {str:str} : external
+thread_sleep: fn float -> void : external
+print: fn *X -> void : external
+
+spy :: fn tag: str, x ->
+    print(tag + " " + as_str(x))
+    ret x
```
FredTheDino, line 12:
```suggestion
    x
```
Don't need this ret

std/math.sy
```diff
@@ -0,0 +1,61 @@
...
+        ret b
+    end
+end
+
+max :: fn a, b ->
+    if a > b do
+        ret a
+    else do
+        ret b
+    end
```
FredTheDino, line 33: Soon...

FredTheDino (approved):
Looks good - just rename the files so I can tell them apart.

sornas (approved):

FredTheDino:
@EMattfolk ping

PR #655, opened 2022-01-21
If expressions - the new expression

### LINKED ISSUE #573: Move to if-expressions
Promote if-statements to be if-expressions!

FredTheDino:
Changes the parsing of if-blocks. If-s are now expressions. The expression is a bit wonky at places and has some edge cases which we probably want to fix (but we'll do that later!).
For example - this is how we write if-else chains now
```
if a do
	x()
else do if z do
		y()
	else do
		p()
	end
end
```
This has to do with if-expressions being changed to expressions. The whole if-thing needs an overhaul TBH.

This snipet of code also doesn't work in the typechecker (I added a failing test and created an issue for it).
```
a :: fn a ->
    if a == 0 do
        ret a
    else do
        ret 1
    end
end
```
The if-expression evaluates to a void-type. The void type is then unified with the returned `int`-s. The nice way to solve this is to have a bottom-type instead of a void. And a bottom can be changed to any other type since it cannot be instantiated. But this is WAY out of scope for this PR.

Closes #573


sylt-compiler/src/typechecker.rs
```diff
@@ -958,48 +939,82 @@ impl TypeChecker {
...
+        ctx: TypeCtx,
+    ) -> TypeResult<(Option<TyID>, Option<TyID>)> {
+        // Left this for Gustav
+        let ss = self.stack.len();
+        let mut ret = None;
+        for stmt in statements.iter_mut() {
+            let stmt = self.statement(stmt, ctx)?;
+            ret = self.unify_option(span, ctx, ret, stmt)?;
         }
+        // We typecheck the last statement twice sometimes, doesn't matte though.
```
EMattfolk, line 1001:
```suggestion
        // We typecheck the last statement twice sometimes, doesn't matter though.
```

EMattfolk (approved):
I don't really like the `else do if`, but it guess it can be fixed later?

PR #659, opened 2022-01-22
case-expressions! (And some parser fixes)

FredTheDino:
Now case-expressions are a thing. The syntax is slightly changed to have `->` in them instead of `do`. I also found a funky parser-bug and with that accidentally imporved the error messages. :shrug: 

There was also a bug with us not type-checking return-statements in case-statements, we do that now aswell. ;)


EMattfolk (approved):
More coverage too!

PR #667, opened 2022-01-24
add exponential format for floats

sornas:


EMattfolk (approved):

sornas:
dunno about forgot but i can add that one too

PR #671, opened 2022-02-02
Simple maybe in the standard library

FredTheDino:
- Simple maybe in the standard library
- Fix minor bug with nested enums
- Add some maybe tests for the std.maybe

### DIFF
--- a/std/maybe.sy
+++ b/std/maybe.sy
@@ -11,7 +11,7 @@ fromMaybe :: fn maybe, a ->
     end
 end
 
-flatten :: fn maybe -> Maybe
+flatten :: fn maybe ->
     case maybe do
         Just x -> x end
         None -> Maybe.None end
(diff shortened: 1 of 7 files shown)


std/maybe.sy
```diff
@@ -0,0 +1,30 @@
+
+Maybe :: enum
+    Just *
+    None
+end
+
+fromMaybe :: fn maybe, a ->
```
iwikal, line 7: `maybe.orDefault` feels like a more descriptive name for this function.

iwikal (approved):

std/maybe.sy
iwikal: I see now that this is stolen from Haskell, while my suggestion is from Rust. I think the Rust name is better though :angel:

Quaqqer: I think it is a lot more intuitive with or as well.

FredTheDino: The Haskell function has the added context of saying what type you give in. I don't personally agree but I am being easily swayed today.

sornas (approved):

Quaqqer (approved):

PR #672, opened 2022-02-02
Don't add returns in the syntax-tree

FredTheDino:
We don't insert returns eagerly in the parser now. We deduce it in the type-checker.
Which gives us a killer new error-message and more control.
(I was gonna fix the if-expression return bug, but we needed to do this first.)

```
  The return-type isn't explicitly set to `void`, but the function returns nothing
```


Quaqqer (approved):
I think it looks good.

sylt-compiler/src/typechecker.rs
```diff
@@ -1119,16 +1119,38 @@ impl TypeChecker {
...
+                    let void = Some(self.push_type(Type::Void));
+                    self.unify_option(span, ctx, actual_ret, void)?
+                };
+
+                if actual_ret.map(|x| self.is_void(x)).unwrap_or(true) && returns_something {
+                    return err_type_error!(
+                        self,
+                        ret.span,
+                        TypeError::Exotic,
+                        "The return type isn't explicitly set to `void`, but nothing is returnd"
```
sornas, line 1139:
```suggestion
                        "The return type isn't explicitly set to `void`, but nothing is returned"
```

sylt-compiler/src/typechecker.rs
```diff
@@ -1119,16 +1119,38 @@ impl TypeChecker {
...
+                    );
+                }
+
+                self.unify_option(span, ctx, Some(ret_ty), actual_ret)
+                    .help(
+                        self,
+                        ret.span,
+                        "The actual return type differ from the specified return type".into(),
```
sornas, line 1147:
```suggestion
                        "The actual return type differs from the specified return type".into(),
```

sornas (approved):

PR #673, opened 2022-02-03
Pure functions

### LINKED ISSUE #658: Pure functions
Pure function definitions, function definitions that disallow:
- Mutations (assignments)
- Reading non-const globals
- Side effects

Criteria:
- 2-character function definition for pures, to be consistent with `fn`. Altenratives are `pu`, `pn`, `la`, `ex`

Quaqqer:
Adds pure functions. Pure functions are normal functions with additional constraints.

An example of a pure function is the following.
```
pure_fn :: pu x ->
  x * 2
end
```

The constraints in a pure function are the following:
- You cannot make assignments.
- You cannot declare mutable variables.
- You cannot read mutable variables (in global scope).

Closes #658.


sylt-compiler/src/typechecker.rs
```diff
@@ -1594,6 +1675,17 @@ impl TypeChecker {
...
+                Constraint::Pure => match self.find_type(a) {
+                    Type::Function(_, _) => Ok(()),
+                    Type::Unknown => Ok(()),
+                    _ => err_type_error!(
+                        self,
+                        span,
+                        TypeError::Impurity,
+                        "Non pure function something something"
+                    ),
+                },
```
Quaqqer, line 1688: I am not actually certain what this code does, hence the bad message. Although I will look in to it.

Quaqqer: Is it that only functions (or unknowns) can have the pure constraint?

FredTheDino: You end up here if you have something that isn't a function, that has the "Pure" tag on it. So if you managed to put in something that cannot be Pure.

Here's an example that should trigger that error:
```
a : fn <*a: Pure> *a -> *a : fn a -> int do a end
```
So your second comment is correct. A better error message might be:
> Only functions can be pure, but this is not a function
Someone else will hve more opinions on the spelling and formulation. But something along those lines.

sylt-compiler/src/typechecker.rs
```diff
@@ -441,6 +448,10 @@ impl TypeChecker {
                 self.add_constraint(var, span, Constraint::Contains(a));
                 self.add_constraint(a, span, Constraint::IsContainedIn(var));
             }
+            "Pure" => {
+                check_constraint_arity(self, span, "Pure", num_args, 0)?;
+                self.add_constraint(var, span, Constraint::Pure);
+            }
```
FredTheDino, line 454: This isn't technically needed right here... But it might be nice to have

sylt-compiler/src/typechecker.rs
```diff
@@ -832,7 +853,18 @@ impl TypeChecker {
...
                         .cloned()
                     {
-                        Some(Name::Global(var)) => no_ret(var.ty),
+                        Some(Name::Global(var)) => {
+                            if ctx.inside_pure && matches!(var.kind, VarKind::Mutable) {
+                                err_type_error!(
+                                    self,
+                                    span,
+                                    TypeError::Impurity,
+                                    "Cannot access mutable variables from pure functions."
```
FredTheDino, line 862: We don't have `.` in our error messages

sylt-compiler/src/typechecker.rs
```diff
@@ -853,6 +885,21 @@ impl TypeChecker {
                                 TypeError::WrongArity { got: args.len(), expected: params.len() }
                             );
                         }
+
+                        if ctx.inside_pure
+                            && !self
+                                .find_node(f)
+                                .constraints
+                                .contains_key(&Constraint::Pure)
```
FredTheDino, line 893: Maybe break this out into a function? We consider `find_node` to be low-level, so it's a good idea to abstract it away. :)

sylt-compiler/src/typechecker.rs
```diff
@@ -853,6 +885,21 @@ impl TypeChecker {
...
+                        {
+                            return err_type_error!(
+                                self,
+                                span,
+                                TypeError::Impurity,
+                                "Cannot call unpure functions from pure functions."
```
FredTheDino, line 899: - .

sylt-parser/src/parser.rs
```diff
@@ -599,10 +599,15 @@ pub fn parse_type<'t>(ctx: Context<'t>) -> ParseResult<'t, Type> {
...
         // Function type
-        T::Fn => {
+        ty @ T::Fn | ty @ T::Pu => {
             let ctx = ctx.skip(1);
 
             let mut constraints = BTreeMap::new();
+
+            if let T::Pu = ty {
+                constraints.insert("Pure".to_string(), Vec::new());
+            }
```
FredTheDino, line 609: Nice!

sylt-compiler/src/typechecker.rs
```diff
@@ -1270,6 +1343,14 @@ impl TypeChecker {
         let span = statement.span;
         let (ident, kind, ty, value) = match &mut statement.kind {
             StatementKind::Definition { ident, kind, ty, value } => {
+                if ctx.inside_pure && !kind.immutable() {
+                    return err_type_error!(
+                        self,
+                        span,
+                        TypeError::Impurity,
+                        "Cannot make mutable declarations in pure functions."
```
FredTheDino, line 1351: - .

sylt-compiler/src/typechecker.rs
```diff
@@ -553,6 +564,16 @@ impl TypeChecker {
 
             StatementKind::Assignment { kind, target, value } => {
                 self.can_assign(span, ctx, target)?;
+
+                if ctx.inside_pure {
+                    return err_type_error!(
+                        self,
+                        span,
+                        TypeError::Exotic,
+                        "Cannot make assignments in pure functions."
```
FredTheDino, line 573: - .

FredTheDino (commented):
Looks like a really good start. I was hoping #672 would be merged by now, but that is apparently a change people are reluctant to read...

Just add some tests for it. Write some simple programs and pop them in the tests folder! Ask if something is unclear!

sornas (commented):
tests would be pog

sylt-compiler/src/typechecker.rs
```diff
@@ -1119,16 +1166,42 @@ impl TypeChecker {
...
+                    .help_no_span(
+                        "The actual return type and the specified return type differ!".into(),
+                    )?;
+
+                if actual_ret.is_none() && returns_something {
+                    return err_type_error!(
+                        self,
+                        ret.span,
+                        TypeError::Exotic,
+                        "The return-type isn't explicitly set to `void`, but the function returns nothing"
```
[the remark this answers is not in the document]
Quaqqer: I guess this should be directed at @FredTheDino but since you suggested it here I changed it.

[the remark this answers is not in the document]
FredTheDino: Technically - this should be part of #672 - but we'll solve it later

Quaqqer:
Design decision time!

These two tests should fail, but for what reason?
https://github.com/sylt-lang/sylt-lang/blob/pure-functions/tests/pure/impure_access_list.sy
https://github.com/sylt-lang/sylt-lang/blob/pure-functions/tests/pure/impure_access_blob.sy

The two options are:
- Disallow reading lists, blobs, etc. that are in scope of the definition because they can be mutated.
- Disallow mutation of lists, blobs, etc. that are not declared as mutable.

FredTheDino:
When it comes to blobs and lists in pure functions, I'd say we allow it now and change how constant works so nothing inside a constant list is mutable and nothing inside a constant blob is mutable. In my opinion this is a better understanding structure.

But the change to constants actually being constants is a little bit to big for what I think this PR should be. It's usually better to merge smaller PRs since they're easier to review. I'd say merge this and fix the mutability as a next PR. How does that feel?

Quaqqer:
> When it comes to blobs and lists in pure functions, I'd say we allow it now and change how constant works so nothing inside a constant list is mutable and nothing inside a constant blob is mutable. In my opinion this is a better understanding structure.

This sounds good.

> But the change to constants actually being constants is a little bit to big for what I think this PR should be. It's usually better to merge smaller PRs since they're easier to review. I'd say merge this and fix the mutability as a next PR. How does that feel?

I agree, my thinking was just that if we wanted to restrict access to blobs, lists, etc. it might be smaller and fit inside this pr.

sylt-compiler/src/typechecker.rs
FredTheDino: It's also a good idea to use the `violating` type error if you have a type that you can point at. That way we print the type that is wrong. ;)

FredTheDino (approved):
Looks good to me!

PR #678, opened 2022-02-06
Pure std

Quaqqer:
Done:
- Add `Pure` constraint to inner type `pu`
- Marked appropriate functions in std/math as pure.

Todo:
- [x] Mark every appropriate std function as pure
- [x] Make sure the function body has the constraints that were defined in the definition. Right now the following should work, although it is incorrect.
      ```
      b := 2
      a : pu -> int : fn -> int
          b
      end
      ```

### DIFF
--- a/std/common.sy
+++ b/std/common.sy
@@ -1,11 +1,8 @@
-dbg :: pu x ->
-    print(x)
-    x
-end
+dbg : pu *X -> *X : external
 
-args: pu -> {str:str} : external
+args: fn -> {str:str} : external
 thread_sleep: fn float -> void : external
-print: pu *X -> void : external
+print: fn *X -> void : external
 
 spy :: pu tag: str, x ->
     dbg(tag + " " + as_str(x))
@@ -17,6 +14,6 @@ split: pu str -> [str] : external
 
 as_float: pu int -> float : external
 as_int: pu float -> int : external
-as_char: pu str -> int : external
+as_char: fn str -> int : external
 as_chars: pu str -> [int] : external
 as_str: pu *X -> str : external
--- a/std/common.sy
+++ b/std/common.sy
@@ -14,6 +14,6 @@ split: pu str -> [str] : external
 
 as_float: pu int -> float : external
 as_int: pu float -> int : external
-as_char: fn str -> int : external
+as_char: pu str -> int : external
 as_chars: pu str -> [int] : external
 as_str: pu *X -> str : external
--- a/std/common.sy
+++ b/std/common.sy
@@ -1,22 +1,22 @@
-dbg :: fn x ->
+dbg :: pu x ->
     print(x)
     x
 end
 
-args: fn -> {str:str} : external
+args: pu -> {str:str} : external
 thread_sleep: fn float -> void : external
-print: fn *X -> void : external
+print: pu *X -> void : external
 
-spy :: fn tag: str, x ->
+spy :: pu tag: str, x ->
     dbg(tag + " " + as_str(x))
     x
 end
 
-unsafe_force: fn *X -> *Y : external
-split: fn str -> [str] : external
+unsafe_force: pu *X -> *Y : external
+split: pu str -> [str] : external
 
-as_float: fn int -> float : external
-as_int: fn float -> int : external
-as_char: fn str -> int : external
-as_chars: fn str -> [int] : external
-as_str: fn *X -> str : external
+as_float: pu int -> float : external
+as_int: pu float -> int : external
+as_char: pu str -> int : external
+as_chars: pu str -> [int] : external
+as_str: pu *X -> str : external
--- a/std/common.sy
+++ b/std/common.sy
@@ -1,3 +1,5 @@
+from maybe use (Maybe)
+
 dbg : pu *X -> *X : external
 
 args: fn -> {str:str} : external
@@ -14,6 +16,6 @@ split: pu str -> [str] : external
 
 as_float: pu int -> float : external
 as_int: pu float -> int : external
-as_char: pu str -> int : external
+as_char: pu str -> Maybe : external
 as_chars: pu str -> [int] : external
 as_str: pu *X -> str : external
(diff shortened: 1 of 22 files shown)


Quaqqer:
I think I start to think it looks okay now, a problem is that you cannot mix pure and impure functions in a list for instance. I don't know how that works in particular, maybe someone could clarify that.

sylt-compiler/src/typechecker.rs
```diff
@@ -483,32 +478,37 @@ impl TypeChecker {
                 return self.type_assignable(ctx, assignable);
             }
 
-            Fn { constraints, params, ret } => {
+            Fn { constraints, params, ret, pure } => {
```
FredTheDino, line 481: Maybe `is_pure`?

sylt-compiler/src/typechecker.rs
```diff
@@ -1804,8 +1796,23 @@ impl TypeChecker {
...
                 }
 
-                (Type::Function(a_args, a_ret), Type::Function(b_args, b_ret)) => {
+                (
+                    Type::Function(a_args, a_ret, a_purity),
+                    Type::Function(b_args, b_ret, b_purity),
+                ) => {
                     // TODO: Make sure there is one place this is checked.
+                    match (a_purity, b_purity) {
+                            (Purity::Undefined, _) |
```
FredTheDino, line 1805: Maybe call it `Unknown`? Maybe? I don't know?

std/math.sy
```diff
@@ -1,59 +1,65 @@
...
 
-reflect: fn (float, float), (float, float) -> (float, float) : external
+reflect: pu (float, float), (float, float) -> (float, float) : external
 
-dot :: fn v1: (float, float), v2: (float, float) -> float
+dot :: pu v1: (float, float), v2: (float, float) -> float
     v1[0] * v2[0] + v1[1] * v2[1]
 end
+
+random : fn -> float : external
```
FredTheDino, line 61: This function is not pure. So it shouldn't be callable inside pure functions, yet it is? :0

FredTheDino (approved):
Lgtm, some variable nameing nits. But it looks pretty solid!

std/math.sy
Quaqqer: It is not, see https://github.com/sylt-lang/sylt-lang/blob/pure-std/tests/pure/impure_std.sy. :P

std/common.sy
```diff
@@ -1,22 +1,22 @@
...
+dbg :: pu x ->
     print(x)
     x
 end
 
-args: fn -> {str:str} : external
+args: pu -> {str:str} : external
 thread_sleep: fn float -> void : external
-print: fn *X -> void : external
+print: pu *X -> void : external
```
sornas, line 8: hmmm

Quaqqer: Oops, hehe

Quaqqer: I should probably look over this some more, I realize now I just put pure everywhere without a second thought.

sornas: honestly it depends on what we consider "pure". is it only pure with respect to variables or is external code allowed to do stuff with the system as long as it isn't visible to sylt-code? one could consider "pure" as something like "const" from rust.

sylt-compiler/src/typechecker.rs
iwikal: Or just have `Some(Pure)`, `Some(Impure)` and `None`. That would allow us to use the same enum to model situations when the potential purity must be known.

std/common.sy
```diff
@@ -17,6 +14,6 @@ split: pu str -> [str] : external
 
 as_float: pu int -> float : external
 as_int: pu float -> int : external
-as_char: pu str -> int : external
+as_char: fn str -> int : external
 as_chars: pu str -> [int] : external
```
sornas, line 18: this feels a bit off

Quaqqer: Hmm, i changed it but a problem is that if the string is empty it actually returns nil rather than an int, so it does not necessarily return an int.

std/common.sy
FredTheDino: I like the idea of not having side-effects at all. So all operations that manipulate a list or change the state or output to the console are considered impure. But we should probably allow `spy` and `dbg` too make debugging eaiser. :D

std/common.sy
FredTheDino: That feels like a problem with the implementation not the purity. If it doesn't modify the string I see this as being pure. I'd rather see we change the return-type of the function to use the Maybe type from the standard library. ;)

std/common.sy
sornas: haskell has https://hackage.haskell.org/package/base-4.16.0.0/docs/Debug-Trace.html for specifically that reason

sylt-compiler/src/typechecker.rs
sornas:
```
pub enum Purity {
	Pure,
	Impure,
}
```
sounds like a `bool`. `struct Pure(bool);`? this becomes¹
```
match (a_purity, b_purity) {
	(None, _) |
	(_, None) |
	(Some(a), Some(b)) if a == b => (),
	// ...
}
```


but i guess this becomes a bit verbose. i'd be happy with `enum Purity { Pure, Impure }` as well but wanted to raise `struct Pure(bool)` just in case. thoughts welcome.

---
¹ although, this could be accomplished with an enum too if we derive Eq

sylt-compiler/src/typechecker.rs
```diff
@@ -1362,14 +1360,19 @@ impl TypeChecker {
...
                     .iter()
                     .map(|_| self.push_type(Type::Unknown))
                     .collect();
                 let ret = self.push_type(Type::Unknown);
-                let fn_ty = self.push_type(Type::Function(args, ret));
+                let purity = if *pure {
+                    Purity::Pure
+                } else {
+                    Purity::Undefined
+                };
```
sornas, line 1374: you know, with a `struct Pure(bool)` this would become

```
let purity = pure.then(|| Pure(true));
```

Quaqqer: I kind of did this, but without the struct. To me `struct Pure(bool)` seems kind of weird.

sornas (approved):

PR #679, opened 2022-02-06
output timing data for chrome://tracing

sornas:
usage:

- build with timing: `$ cargo build --features timed`.
- run and redirect output: `$ target/debug/sylt <file.sy> --trace trace.json`.
- open a chrome-based browser (brave, chromium, ...), go to `chrome://tracing`
  and `load` `trace.json`.

(you probably want to use release mode though)

this is Good Enough:tm: for the big picture, but since every timing has some overhead it isn't equivalent to using a profiler, especially on the smaller scale or when there are a lot of calls.

things to do:

- [x] time blocks, not only functions
- [ ] use pid/tid?
- [x] (in progress) use args? maybe so we can see e.g. what types are taking a long time to
  parse.
- [ ] use categories?
- [x] leave eprintln for actual errors and dump the json directly
- [x] conditional in runtime as well so we can build with timing but don't _have_
  to output timing data.
- [ ] [better macro error reporting](https://github.com/dtolnay/syn/blob/master/examples/lazy-static/lazy-static/src/lib.rs)
- [ ] dump data without requiring program to terminate
- [ ] less overhead?


sornas:
feeling happy enough now. i think these can be left as issues

- [better macro error reporting overall](https://github.com/dtolnay/syn/blob/master/examples/lazy-static/lazy-static/src/lib.rs?rgh-link-date=2022-02-06T18%3A43%3A48Z)
- use categories to better filter tracing
- print/save the json data without requiring the compiler to terminate

sylt/src/lib.rs
```diff
@@ -95,9 +98,16 @@ pub struct Args {
         long = "output",
         short = "o",
         meta = "FILE",
-        help = "Output a compiled lua file, '-' for stdout"
+        help = "Output a compiled lua file. '-' for stdout."
```
FredTheDino, line 101: Very agressive ;)

sylt/src/main.rs
```diff
@@ -24,6 +28,21 @@ fn main() -> Result<(), String> {
...
+        let output = sylt_macro::timed_trace!(sylt, sylt_compiler, sylt_parser);
+
+        if outfile == &std::path::Path::new("-") {
+            eprintln!("{}", output,);
+        } else {
+            if let Err(e) = std::fs::write(outfile, output)
+                .map_err(|e| format!("failed to write trace to file: {}", e))
+            {
+                eprintln!("{}", e);
+            }
```
FredTheDino, line 42: Can't the format be placed in the eprintln! directly?

FredTheDino (approved):
LGTM

FredTheDino:
@sornas  Progress on this? It looks like the merge diff is only getting bigger TBH.

sornas:
I think some handles were lost in the merge but it's easy to add them back.

PR #680, opened 2022-02-08
elif

FredTheDino:
We now have `elif`, so if-else chains become a lot nicer. This also simplified both the typechecker and the parser. Which is nice.

The new syntax is more flexible and less verbose, something like this:

```
if true do
    aaa
elif false do
    bbb
else
    ccc
end
```

The typechecker now recognizes these statements are "void" statements:
```
if true do
    aaa
elif false do
    bbb
end
```
So if you don't cover all cases nothing is returned here. (Ofcourse it's only the typechecker, hehehe)

### DIFF
--- a/sylt-compiler/src/intermediate.rs
+++ b/sylt-compiler/src/intermediate.rs
@@ -337,29 +337,17 @@ impl<'a> IRCodeGen<'a> {
                         if let Some(cond) = condition {
                             let (expr, var) = self.expression(&cond, ctx);
                             let block = self.expression_block(out, body.clone(), ctx);
-                            [
-                                expr,
-                                vec![IR::If(var)],
-                                block,
-                                vec![IR::Else],
-                            ].concat()
+                            [expr, vec![IR::If(var)], block, vec![IR::Else]].concat()
                         } else {
                             let var = self.var();
                             let block = self.expression_block(out, body.clone(), ctx);
-                            [
-                                vec![IR::Bool(var, true),IR::If(var)],
-                                block,
-                            ].concat()
+                            [vec![IR::Bool(var, true), IR::If(var)], block].concat()
                         }
                     })
                     .flatten()
                     .collect::<Vec<_>>();
                 (
-                    [
-                        code,
-                        branches.iter().map(|_| IR::End).collect()
-                    ]
-                    .concat(),
+                    [code, branches.iter().map(|_| IR::End).collect()].concat(),
                     out,
                 )
             }
--- a/sylt-compiler/src/intermediate.rs
+++ b/sylt-compiler/src/intermediate.rs
@@ -344,7 +344,12 @@ impl<'a> IRCodeGen<'a> {
                                 vec![IR::Else],
                             ].concat()
                         } else {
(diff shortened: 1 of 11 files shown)


sylt-compiler/src/intermediate.rs
```diff
@@ -329,21 +329,25 @@ impl<'a> IRCodeGen<'a> {
...
+                    .map(|IfBranch { condition, body, span: _ }| {
+                        if let Some(cond) = condition {
+                            let (expr, var) = self.expression(&cond, ctx);
+                            let block = self.expression_block(out, body.clone(), ctx);
+                            [expr, vec![IR::If(var)], block, vec![IR::Else]].concat()
+                        } else {
+                            let var = self.var();
+                            let block = self.expression_block(out, body.clone(), ctx);
+                            [vec![IR::Bool(var, true), IR::If(var)], block].concat()
+                        }
```
Quaqqer, line 345: Maybe it would be good to make sure that only the last branch is missing a condition, or is that just unnecessary work?

sylt-compiler/src/typechecker.rs
```diff
@@ -1130,17 +1130,57 @@ impl TypeChecker {
...
+                    Some(void)
+                } else {
+                    let mut value = None;
+                    for (span, branch_ret, branch_value) in tys.iter() {
+                        // TODO(ed): These are bad errors, they're easy to confuse. A better
+                        // formulation?
+                        ret = self
+                            .unify_option(*span, ctx, *branch_ret, ret)
+                            .help_no_span(
+                                "The return from this block doesn't match the other branches"
```
Quaqqer, line 1173: earlier branches instead, since it could be an earlier branch that is of the wrong type?

tests/if_else/do_no_path.sy
```diff
@@ -0,0 +1,14 @@
...
+        <!>
+    end
+end
+
+start :: fn do
+    f' 3
+    f' 0
+end
+
+// error: type isn't explicitly set to `void`
```
Quaqqer, line 14: Not related to this pr, but...

Isn't this a bit weird. The function is declared with `f :: fn a do` and not `f :: fn a ->`, and since there is no need for the ret keyword you cannot know at a glance if the function actually returns something. Take this for instance:

```
g :: fn do
    1
end

f :: fn do
    g'
end

start :: fn do
    print' f'
end
```

If you look at the definition of `f` there is no clear indication that it actually returns a value rather than void.

Quaqqer (commented):
Overall I think it looks good!

sylt-compiler/src/intermediate.rs
FredTheDino: When we parse it we parse else-branches differntly and always at the end. The parsing code looks a bit like this:
```
parse_if()
while is_elif(): parse_elif()
if is_else(): parse_else()
```
So it *shouldn't* be possible. I was wondering a bit if I made the IfBranch struct to open, but I didn't really see how we would get anything valuable out of labeling the branches with a custom enum instead of an option.

I don't see this as a huge problem. But feel free to come discuss another way of doing it.

EMattfolk (approved):

PR #682, opened 2022-02-13
Syntax for generics

FredTheDino:
Added syntax for generics, this is now a valid program:

```
A :: enum(*A, *B)
    C *A,
    D *B,
end

B :: blob(*A, *A) {
    a: *A,
    b: *A,
}

start :: fn do
    a : A(int, int) : A.C 1
    b : B(int) : B { a: 1, b: 2 }
end
```

I also did some quite agressive changes to propogate spans in the type-checker.
We can now point a lot better att wierd generic errors. Maybe it shouldn't have
been in this PR, but I just happened to do it.

It's not quite done yet, but I'd rather get feedback now if something is extreemly off.
If you find a good test-case while testing add it somewhere so we can  add it to the test-suite. :D

### DIFF
--- a/sylt-parser/src/parser.rs
+++ b/sylt-parser/src/parser.rs
@@ -95,6 +95,18 @@ impl Identifier {
     }
 }
 
+impl PartialOrd for Identifier {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.name.cmp(&other.name))
+    }
+}
+
+impl Ord for Identifier {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.name.cmp(&other.name)
+    }
+}
+
 impl PartialEq for Identifier {
     fn eq(&self, other: &Self) -> bool {
         self.name == other.name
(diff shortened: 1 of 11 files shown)


tests/hm_typing/simple_generics.sy
```diff
@@ -0,0 +1,14 @@
...
+end
+
+B :: blob(*A, *B) {
+    a: *A,
+    b: *B,
+}
+
+start :: fn do
+    a : A(int, int) : A.C 1
+    b : B(int) : B { a: 1, b: 2 }
```
sornas, line 13: it seems like both A and B is int here. what happens if i specify 2 out of 3 generics?

sylt-parser/src/statement.rs
```diff
@@ -427,71 +429,101 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
...
+                        let ctx = ctx.skip_if(T::Colon);
+                        let (ctx, ty) = parse_type(ctx)?;
+                        if !matches!(ctx.token(), T::Comma | T::End | T::Newline) {
+                            raise_syntax_error!(ctx, "Expected a delimiter ','");
+                        };
+                        (ctx, ty)
+                    };
+                    Ok((ctx.skip_if(T::Comma), (variant, ty)))
                 }
+                parse_sep_end_by(ctx, &sep, &end, &item)?
```
sornas, line 508: i _think_ function pointers are automatically borrowed

sylt-parser/src/parser.rs
```diff
@@ -384,6 +402,54 @@ impl<'a> Context<'a> {
     }
 }
 
+fn parse_sep_end_by<'t, Item>(
+    ctx: Context<'t>,
+    sep: &dyn Fn(Context<'t>) -> ParseResult<'t, ()>,
+    end: &dyn Fn(Context<'t>) -> ParseResult<'t, bool>,
+    item: &dyn Fn(Context<'t>) -> ParseResult<'t, Item>,
```
sornas, line 409: &dyn -> impl

sylt-parser/src/statement.rs
FredTheDino: I had to add it. Tried to remove it.

FredTheDino: I think impl solved this too :o:

tests/hm_typing/simple_generics.sy
sornas: reasonable

sornas (approved):

PR #688, opened 2022-03-04
Let's add another syntaxtree!

FredTheDino:
I heard you like syntax trees, so I translated the syntax tree to a different syntax tree, with a syntax tree.

We now have a seperate step for resolving the names of variables. This step
generates a new syntax tree and a list of variables. The idea is to propogate
these changes further down the line into the dependency checker, type checker
and then code generation.

This should remove some complexity and fix the nasty bug we have where we
codegen and typecheck different values.


sylt-compiler/src/name_resolution.rs
```diff
@@ -0,0 +1,1038 @@
...
+                        self.push_var(ident.clone(), *kind);
+                        self.expression(value)?
+                    }
+                    (false, false) => {
+                        // Value, push the var after!
+                        let expr = self.expression(value)?;
+                        self.push_var(ident.clone(), *kind);
+                        expr
+                    }
+                };
```
iwikal, line 765: Kind of a confusing match on `(bool, bool)` imo.
```suggestion
                let value = if matches!(
                    value.kind,
                    sylt_parser::ExpressionKind::Function { .. },
                ) {
                    // Function, push the var before!
                    self.push_var(ident.clone(), *kind);
                    self.expression(value)?
                } else if self.stack.is_empty() {
                    // Outer statement - it's a global so just evaluate the value!
                    self.expression(value)?
                } else {
                    // Value, push the var after!
                    let expr = self.expression(value)?;
                    self.push_var(ident.clone(), *kind);
                    expr
                };
```

sylt-compiler/src/name_resolution.rs
```diff
@@ -0,0 +1,1038 @@
...
+        Ok(match &ty.kind {
+            TK::Implied => T::Implied(span),
+            TK::Resolved(t) => T::Resolved(t.clone(), span),
+            TK::UserDefined(t, gs) => {
+                let r = match self.ty_assignable(t)? {
+                    T::UserType(r, _) => r,
+                    _ => raise_resolution_error! {
+                        self,
+                        span,
+                        "This is not a reference to a user defined type"
```
iwikal, line 435: I think this is too much information, it would probably suffice to say "This is not a type". They might have misspelled a built-in type, so the desired type needn't be user defined.
```sylt
let x: boolean : true
```

iwikal: Upon further reading, this looks like dead code? `ty_assignable` only ever returns `UserType` as far as I can tell.

sylt-compiler/src/name_resolution.rs
```diff
@@ -0,0 +1,1038 @@
...
+                    },
+                };
+                let gs = self.type_vec(gs)?;
+                T::UserDefinedWithGenerics(r, gs, span)
```
iwikal, line 439: So if I'm understanding this correctly, `UserDefinedWithGenerics` is used even if `gs` is empty? Then what is `UserType` for? Is there a need for both?

sylt-compiler/src/name_resolution.rs
```diff
@@ -0,0 +1,1038 @@
...
+            Some(Name::Name(_)) => raise_resolution_error! {
+                self,
+                ident.span,
+                "{:?} is a variable, not a type",
+                ident.name
+            },
+            None => raise_resolution_error! {
+                self,
+                ident.span,
+                "{:?} no type named",
```
iwikal, line 482: I don't really understand this error message.

sylt-compiler/src/name_resolution.rs
```diff
@@ -0,0 +1,1038 @@
...
+                Type::UserType(self.lookup(span.file_id, &ident.name, ident.span)?, span)
+            }
+            TAK::Access(namespace, ty) => {
+                let new_namespace = self.namespace_type_list(span.file_id, &namespace)?;
+                match self.lookup_global(new_namespace, &ty.name) {
+                    Some(Name::Name(r)) => Type::UserType(*r, span),
+                    None => raise_resolution_error! {
+                        self,
+                        ty.span,
+                        "{:?} no type named",
```
iwikal, line 503: I don't really understand this error message.

iwikal (commented):
Not a super thorough review, I haven't looked much at the logic, but we have tests so :shrug: I'm sure it's fine!

sylt-compiler/src/name_resolution.rs
iwikal: Also, it looks like global functions get pushed even though they should already be handled in a previous pass? Is this intentional?

sylt-compiler/src/name_resolution.rs
FredTheDino: In some places you don't want to specify the generics (or maybe you do, but we don't allow it right now). The prime usecase is when instantiating a blob. We write something along the lines of:
`A.B.Blob { ... }`. I'll merge them. Nice catch!

sylt-compiler/src/name_resolution.rs
FredTheDino: This is the only piece of complicated logic in the program, and your translation won't match the behaviour of my match-block. And I've recently convinced myself my matchblock is wrong and your if-statements are incorrect. But it might be more readable as an if-else chain?

And you're point that `global functions get pushed` are correct, this is not the behaviour we want. We want local variables to live on the stack. Or maybe I've thought about this, because we need to push something to show that we're inside a function. Maybe we should push a dummy value for entering a function - or setting a flag. At this point I'm rambling. But yes, this is wrong.

sylt-compiler/src/name_resolution.rs
FredTheDino: I was kinda safeguarding against future changes. Right now an `unreachable` would be correct, but it might backfire later. I'm maybe being too defensive though.

But I explicitly wanted it to say that it wasn't a user defined type. So I'm still leaning towards keeping the error message. Since it would still be an error if the type was written as `int`. But I might be rambling even more right now.

Quaqqer (approved):

PR #689, opened 2022-03-06
Fix the rest of the compilation pipeline!

FredTheDino:
We now use the resolved variables. It works great!

### DIFF
--- a/sylt-compiler/src/compiler.rs
+++ b/sylt-compiler/src/compiler.rs
@@ -1,5 +1,3 @@
-#![allow(unused)]
-use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::io::Write;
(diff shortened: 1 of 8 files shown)


sylt-compiler/src/compiler.rs
```diff
@@ -70,31 +61,29 @@ impl Compiler {
     }
 
     #[cfg_attr(timed, sylt_macro::timed("compile"))]
-    fn compile(mut self, lua_file: &mut dyn Write, tree: AST) -> Result<(), Vec<Error>> {
+    fn compile(&mut self, lua_file: &mut dyn Write, tree: AST) -> Result<(), Vec<Error>> {
```
FredTheDino, line 64: Does the '&' do anything?

PR #692, opened 2022-04-03
Fixup Jordgubb so it works with the new version of the language

FredTheDino:


Quaqqer (approved):
Looks good.

Only comment I have is that it would be nice if externblobs were used.

PR #693, opened 2022-04-03
Clean up the compiler flags

FredTheDino:
Add the `--no-std` flag, usefull when debugging imports and the typechecker since it dramatically reduces the size of the program.
Also added the `--require/-r` flag, which lets you require a lua file. This is usefull if you want to inject some lua code (like a library written in lua). The reasoning for having only one is so you collect all the imports in one file. I think it makes the most sense.

### DIFF
--- a/sylt/src/lib.rs
+++ b/sylt/src/lib.rs
@@ -23,11 +23,11 @@ where
     R: Fn(&Path) -> Result<String, Error>,
 {
     let file = PathBuf::from(args.args.first().expect("No file to run"));
-    let tree = sylt_parser::tree(&file, reader, true)?;
+    let tree = sylt_parser::tree(&file, reader, !args.no_std)?;
     if args.dump_tree {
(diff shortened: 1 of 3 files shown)


sylt/src/lib.rs
```diff
@@ -97,14 +97,19 @@ pub struct Args {
...
-    #[options(short = "v", no_long, count, help = "Increase verbosity (max 2)")]
-    pub verbosity: u32,
-
     #[options(
-        long = "format",
-        help = "Format the file and write the result to stdout"
+        long = "require",
+        short = "r",
+        meta = "FILE",
+        help = "A lua file to require, usefull for external code. The argument is placed directly in a \"require\""
```
Quaqqer, line 104:
```suggestion
        help = "A lua file to require, useful for external code. The argument is placed directly in a \"require\""
```

Quaqqer (approved):
Looks good

Issue #611, opened 2021-11-28
Dicts do not have out of bounds guards


EMattfolk:
```
start :: fn do
    {"a": fn do end}["b"]()
end
```

EMattfolk:
Do we have some way of testing runtime errors?

FredTheDino:
The index should return a maybe type - so the signature is all wrong TBH. We should add Maybe to STL so we can fix it :D

Issue #536, opened 2021-11-06
Fix `in` for dicts


FredTheDino:
This should work:
```
    a :: {0: 1}
    (0 in a) <=> false
```

Also enable tests:
 - tests/dict/_contains.sy
 - tests/dict/_faulty_contains.sy

PR #695, opened 2022-04-15
Add better sets and dicts

FredTheDino:
Added a set and dict-type written in Lua/Sylt. These new types work for tuples, unlike the previous silly ones. I'm currently thinking along the lines of removing Sets and Dicts from the language as constructs.


std/dict.sy
```diff
@@ -0,0 +1,32 @@
...
+Dict :: blob(*K, *V) {
+    __PHANTOM_KEY: *K,
+    __PHANTOM_VALUE: *V
+}
+
+dict_new : pu<K: CmpEqu> -> Dict(*K, *V) : external
+dict_update : fn<K: CmpEqu> Dict(*K, *V), *K, *V -> void : external
+dict_remove : fn<K: CmpEqu> Dict(*K, *V), *K -> void : external
+
+// We have to sacrefice purity for this to work in Sylt - so I wrote it in Lua.
```
EMattfolk, line 12:
```suggestion
// We have to sacrifice purity for this to work in Sylt, so I wrote it in Lua.
```

EMattfolk (commented):
It may be worth taking a look at #611 and #536 while at it

PR #699, opened 2022-04-19
remove dicts and sets

FredTheDino:
- Death to dict and set
- Fix the tests


Quaqqer (commented):
Looks good I think, although I have two opinions.

- Maybe it would be preferable to keep the syntax for sets and dictionaries and use them as syntactic sugar for the more verbose set.from_list.
- We could add syntax for `... : ... : external "..."` to avoid cluttering the namespace, a negative aspect would be that code with side effects could be run this way, but another benefit is allowing loading external libraries with `require(...)`.

What do you think?

FredTheDino:
> Looks good I think, although I have two opinions.
> 
>     * It would be preferable to keep the syntax for sets and dictionaries and use them as syntactic sugar for the more verbose `set.from_list`.
> 
>     * We could add syntax for `... : ... : external "..."` to avoid cluttering the global lua namespace, a negative aspect would be that code with side effects could be run this way, but another benefit is allowing loading external libraries with `require(...)`.
> 
> 
> What do you think?

Syntax suger is another PR. I don't see a lot of value in adding it but if someone wants to do it - go ahead. It will be messy to test since it wouldn't compile without the standard library - something that makes it a lot easier to troubel shoot in the typechecker. I also don't see it being a lot of value. But still - another PR.

I'm liking the `external "lua-code"` idea better and better. Just make sure it's only available in the top scope. Using that `external` we could inline the Lua functions directly - which would make it a lot easier to read and edit the code. But yet again - this is another PR.

Quaqqer (approved):
> > Looks good I think, although I have two opinions.
> > ```
> > * It would be preferable to keep the syntax for sets and dictionaries and use them as syntactic sugar for the more verbose `set.from_list`.
> > 
> > * We could add syntax for `... : ... : external "..."` to avoid cluttering the global lua namespace, a negative aspect would be that code with side effects could be run this way, but another benefit is allowing loading external libraries with `require(...)`.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you think?
> 
> Syntax suger is another PR. I don't see a lot of value in adding it but if someone wants to do it - go ahead. It will be messy to test since it wouldn't compile without the standard library - something that makes it a lot easier to troubel shoot in the typechecker. I also don't see it being a lot of value. But still - another PR.
> 
> I'm liking the `external "lua-code"` idea better and better. Just make sure it's only available in the top scope. Using that `external` we could inline the Lua functions directly - which would make it a lot easier to read and edit the code. But yet again - this is another PR. Though the `require` thing is technically already fixed. But you are limited to ONE lua file. But this lua file can ofcourse include requires.

I agree, I just think that it makes sense to consider these things before removing the code that parses sets and dicts.

Another cool thing with `external`, depending on your viewpoint, is that it would make it possible to define lua functions directly from sylt.

PR #700, opened 2022-04-22
So... This happened...

FredTheDino:
List indexing isn't a thing anymore! :D

> Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away.

― Antoine de Saint-Exupéry


EMattfolk (approved):

sornas:
i would prefer `b.get(0)` (with your chosen token instead of `.` if required) but it's not a blocker.

sornas (approved):

PR #703, opened 2022-05-27
Lua Definitions

Quaqqer:
# Allow lua definitions in Sylt code.

Examples:

```
a : int : `return 1`
```

```
a : bool : `
    local a = 0
    local b = 1
    return a < b
`
```

## Bugs

- Only works as an outer statement. This is a bug in the current implementation. Maybe it
  should only work as an outer statement however?

## Thoughts

- Beginning with `lua` and ending with `luaend` is cleaner, I think, but it
  might be hard to do since `logos` does not support look-around or non-greedy
  matching in regexes (from what I can find anyway).
- Should inline lua be in functions? It is nice because you get a more narrow
  scope and you can do side effects cleanly, kind of? The fact that only
  lua definitions are allowed (in sylt) makes side effects kinda weird in this case.
- Should it be an expression to allow for lua side effects
  in sylt functions?
- Should the lua be parsed to check for errors?

# Other

- Allow escaping strings with `\"`, like: `" -> \" <- "`
- Allow multi line strings in sylt, before it would all work except for the
  generated lua code.

### DIFF
--- a/sylt-compiler/src/lua.rs
+++ b/sylt-compiler/src/lua.rs
@@ -174,6 +174,13 @@ impl<'a, 'b> Generator<'a, 'b> {
                     write!(self.out, e);
                 }
 
+                IR::Lua(t, lua) => {
+                    let t = self.expand(t);
+                    write!(self.out, "{}", t);
+                    write!(self.out, " = ");
+                    write!(self.out, "(function () {} end)()", lua);
+                }
+
                 IR::Call(t, f, args) => {
                     write!(self.out, "local ");
                     let t = self.expand(t);
--- a/sylt-compiler/src/lua.rs
+++ b/sylt-compiler/src/lua.rs
@@ -115,7 +115,9 @@ impl<'a, 'b> Generator<'a, 'b> {
 
                 IR::Neg(t, a) => ii!(self, t, "(-{})", a),
 
-                IR::Str(t, s) => iis!(self, t, "\"{}\"", s),
+                // Replace newlines in strings with \n in lua, since lua strings don't continue on
+                // the next line
+                IR::Str(t, s) => iis!(self, t, "\"{}\"", s.replace("\n", r"\n")),
                 IR::Float(t, f) => iis!(self, t, "{:?}", f),
 
                 IR::Equals(t, a, b) => ii!(self, t, "({} == {})", a, b),
--- a/sylt-compiler/src/typechecker.rs
+++ b/sylt-compiler/src/typechecker.rs
@@ -541,6 +541,8 @@ impl TypeChecker {
                 }
             }
 
+            S::LuaDefinition { var, ty, span, .. } => self.external_definition(ctx, var, ty, span),
+
             S::Unreachable(_) => Ok(None),
 
             S::Blob { .. } | S::Enum { .. } | S::ExternalDefinition { .. } => {
@@ -644,8 +646,11 @@ impl TypeChecker {
             }
 
             S::ExternalDefinition { var, ty, span, .. } => {
-                let ty = self.resolve_type(ctx, ty)?;
-                self.unify(*span, ctx, self.variables[*var].ty, ty)?;
+                self.external_definition(ctx, var, ty, span)?;
+            }
+
+            S::LuaDefinition { var, ty, span, .. } => {
+                self.external_definition(ctx, var, ty, span)?;
             }
 
             S::Assignment { .. }
@@ -2047,6 +2052,17 @@ impl TypeChecker {
 
     fn span_file(&self, span: &Span) -> FileOrLib {
         self.namespace_to_file[&span.file_id].clone()
+    }
+
+    fn external_definition(
+        &mut self,
+        ctx: TypeCtx,
+        var: &usize,
+        ty: &ResolverType,
+        span: &Span,
+    ) -> TypeResult<Option<TyID>> {
+        let ty = self.resolve_type(ctx, ty)?;
+        Ok(Some(self.unify(*span, ctx, self.variables[*var].ty, ty)?))
     }
 }
 
--- a/sylt-compiler/src/typechecker.rs
+++ b/sylt-compiler/src/typechecker.rs
@@ -541,11 +541,12 @@ impl TypeChecker {
                 }
             }
 
-            S::LuaDefinition { var, ty, span, .. } => self.external_definition(ctx, var, ty, span),
-
             S::Unreachable(_) => Ok(None),
 
-            S::Blob { .. } | S::Enum { .. } | S::ExternalDefinition { .. } => {
+            S::Blob { .. }
+            | S::Enum { .. }
+            | S::ExternalDefinition { .. }
+            | S::LuaDefinition { .. } => {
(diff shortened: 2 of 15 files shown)


sylt-compiler/src/lua.rs
```diff
@@ -174,6 +176,13 @@ impl<'a, 'b> Generator<'a, 'b> {
                     write!(self.out, e);
                 }
 
+                IR::Lua(t, lua) => {
+                    let t = self.expand(t);
+                    write!(self.out, "{}", t);
+                    write!(self.out, " = ");
+                    write!(self.out, "(function () {} end)()", lua);
```
FredTheDino, line 183: Nice! This feels like the best way to do this. Very clean!

tests/lua/string.sy
```diff
@@ -0,0 +1,16 @@
+lua1 : str : `return "quotation: \" <-"`
```
FredTheDino, line 1: Nice test!

tests/lua/basic.sy
```diff
@@ -0,0 +1,5 @@
+a : int : `return 1`
```
FredTheDino, line 1: These tests make me wonder if the return should be implicit? That way you write a lua expression, which might feel more natural?

sylt-compiler/src/typechecker.rs
```diff
@@ -644,8 +646,11 @@ impl TypeChecker {
             }
 
             S::ExternalDefinition { var, ty, span, .. } => {
-                let ty = self.resolve_type(ctx, ty)?;
-                self.unify(*span, ctx, self.variables[*var].ty, ty)?;
+                self.external_definition(ctx, var, ty, span)?;
+            }
+
+            S::LuaDefinition { var, ty, span, .. } => {
+                self.external_definition(ctx, var, ty, span)?;
```
FredTheDino: Alternatively you could have combined the cases - it's fine this way though. But I'm not sure the function warrents its own existence.

sylt-parser/src/statement.rs
```diff
@@ -113,6 +113,23 @@ pub enum StatementKind {
         ty: Type,
     },
 
+    /// Defines a variable with result of lua code (wrapped in a function).
+    ///
+    /// Example:
+    /// ```
+    /// v: void : lua
```
FredTheDino: This isnt the current syntax, right?

sylt-parser/src/statement.rs
```diff
@@ -671,6 +688,11 @@ pub fn statement<'t>(ctx: Context<'t>) -> ParseResult<'t, Statement> {
 
             if matches!(ctx.token(), T::External) {
```
FredTheDino: Consider making this a `match`? I might be nit picking too much though - I'm fine with the if-else.

FredTheDino:
Some thoughts:
- we should parse the lua to make sure it's valid, we already depend on Lua being installed so not a huge step. Can absolutely be a seperate PR. 
- isn't it a normal rust regex? Have you tried the `?`-operator? https://stackoverflow.com/questions/11898998/how-can-i-write-a-regex-which-matches-non-greedy#11899069
- i only want to allow lua code in the global scope. It's a bit wierd how this would work otherwise since there are side effects.
 - inline lua should be in functions. But maybe it should always be an expression?

FredTheDino (commented):
I wanna run this too. But other than that this looks great! Good work!

Quaqqer:
> - isn't it a normal rust regex? Have you tried the `?`-operator?

It is not in fact, `logos` does not depend on regex and does not allow lookaround or non-greedy matching. The reasoning beeing REDOS, denial of service with heavy regexes which would make it not blazing fast. (Which I think is a bad reason for not supporting it.)

> This is doubly problematic since Logos doesn't support non-greedy matching ([for good reasons](https://en.wikipedia.org/wiki/ReDoS)). [1]

[1]: https://github.com/maciejhirsz/logos/issues/133

> - i only want to allow lua code in the global scope. It's a bit wierd how this would work otherwise since there are side effects.

Makes sense

>  - inline lua should be in functions. But maybe it should always be an expression?

It would make sense. I do think the fact that inline lua is (in its current state) of the same structure as a lua module is kinda cool though. :P

An added bonus is that you can write stuff like
```
Requests :: externblob {
    post: fn str -> void,
    get: fn str -> void,
}

requests :  : `
    local r = require('requests')

    return {
        get = r.get
        post = r.post
    }
`
```

Quaqqer:
I think some documentation is missing as well, I should add that.

sylt-compiler/src/typechecker.rs
Quaqqer: I made it a function because I wanted inline lua to work as inner statements, but it doesn't work there anyway...

FredTheDino:
> This is doubly problematic since Logos doesn't support non-greedy matching ([for good reasons](https://en.wikipedia.org/wiki/ReDoS)). https://github.com/maciejhirsz/logos/issues/133

This mentions:
`/\*([^*]|\*+[^*/])*\*+/` which handles a different opening and closing symbol. I do not see the problem here, so I might be missing something. But the regexes they suppose in the thread look functional to me. I'm not entirely convinced this is impossible because of the greedy-ness off the parser.

> It would make sense. I do think the fact that inline lua is (in its current state) of the same structure as a lua module is kinda cool though. :P

Absolutely! Keep the explicit return, very good point!

> I think some documentation is missing as well, I should add that.

It's missing everywhere. But I honeslty don't know how valuable documentation is. I haven't seen that we get much millage out of the time spent there. But if you want to document the code or update the code examples - go for it!

There are also some things which I think we can remove now that we have inline lua.
 - `external`
 - the `require` flag can also be removed
 - We should also rework the entire std-lib, it's gonna be great!

FredTheDino:
These error messages are very off. Maybe parse the ` token so we don't fail if we have an odd number of them?

```
a : bool : `
    local a = true
    return true
`

b : bool : `
    local x = true
    `
    return x
`

start :: fn do
    print' a
end
```

Yields:
```
syntax error: a.sy:4
      Syntax Error on line 4
      Expected newline to end statement
   2 |     local a = true
   3 |     return true
   4 | `
                  ^

syntax error: a.sy:5
      Syntax Error on line 5
      No valid expression starts with 'Error'
   3 |     return true
   4 | `
   5 |
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
I don't know how hard this problem is to solve. It can wait til another PR, but maybe add in the test as a faulty test so we see when it changes?


This program also causes problems:
```
a : fn -> void : `
    return function() print(10 + 10) end
`

start :: fn do
    print' a'
end
```

Yields:
```
typecheck error: a.sy:4
      The `void` has no values and cannot be put in a variable
   2 |     return function() print(10 + 10) end
   3 | `
   4 |
                   ^
help:
      Requirement came from
   2 |     return function() print(10 + 10) end
   3 | `
   4 |
                   ^

Error: "1 errors occured."
```

It kinda looks like the line numbers are off because of the code block. Which is pretty annoying when you're trying to debug programs.


Have a look at these things and then let me know and we'll merge it! :D

PR #705, opened 2023-02-09
sylt the new flavour

FredTheDino:
# The new version of Sylt-Lang
The previous version of Sylt-Lang was really nice, I liked it. But I wanted to push the language further than I think it could go. These are some things I want to be able to achive:
 - Compile to C or something closer to machine code. Sylt was very C like, but
   the way generics worked made it hard to compile to C-code since different
   instances of a function would need different definitions of e.g. Addition.
   There was also some slowness from the overloading and closures were also
   pretty tricky.

 - Complete insight from the compiler into the ongoing in the code and an easy
   way to reason about the code. The old version of Sylt had mutability insight
   it, mutability itself isn't a problem, but it made it really hard for the
   compiler to reason about certain pieces of code, now that everything is
   immutable (unless you cheat) we can make aggressive optimizations on the AST
   and simplify code-gen a great deal.

 - The sylt language was a bit to big for my taste and the syntax was a bit
   cobbled together. This made the parsing kinda hard. The throughput of this
   compiler is on my machine around 500MB of source code per second if I've
   counted right, but I need to check this further.
   
 - The old type checker required exponential amounts of memory, this is not the case with the new version.

# Where is sylt going?
I really want to move towards integrating the game engine and the compiler. The
easiest way to do this is with something that supports hot-code reloading, and
my current idea is to use Lua to make this easy as pie. Some ideas include
functions for tweakable constants in the code, a game-centric API, and saving
each game state and input each frame for bug reproducability.

I also realized that crashes aren't that big a problem. I aim to add functions
that allow you to crash from anywhere, then the meta program in the game engine
can recover this crash. In prod it might drop an input or something to not
crash, but when developing we can probably make something more useful.

# What needs doing?
If you feel like you want to help out, let me know! Right now, the most helpful
is probably using (and abusing) the language - which can generate more tests.
The syntax of the language is quite stable according to me. It should also be
dead simple to integrate language servers and such with this language, and a
formatted would be greatly appreciated (even though it's probably quite hard to
write one).


Quaqqer:
I've begun looking through the code. Mostly learning the new language, documenting the ast module as well as testing some code, https://github.com/sylt-lang/sylt-lang/tree/im-gonna-dooocument.

Comments so far:
- Is `end` in match expressions necessary? You could always use parenthesis instead.
- Foreign could be normal expressions, at least if the syntax (foreign -[[1]]- : Int) was available. Maybe not very useful, I'm undecided.

I also found a bug:

```sylt
def main ::= 
    let x = getX ' { x: 2, y: 3 }
    in x
def getX :: { x } = x
```
--> 
```
lua: stdin:66: attempt to call global '_1_getX' (a nil value)
stack traceback:
        stdin:66: in function <stdin:65>
        stdin:65: in main chunk
        [C]: in ?
```

FredTheDino:
> Is end in match expressions necessary? You could always use parenthesis instead.

It's needed if we want to distinguish between nested matches. Consider this piece of code:
```
match 1
with 1 -> match 2
          with 2 -> amazing
    with 3 -> where_does_this_go
```

> Foreign could be normal expressions, at least if the syntax (foreign -[[1]]- : Int) was available. Maybe not very useful, I'm undecided.

Yes, this could be done, but it also kinda nudges you into using lua everywhere. I deliberately avoided this to nudge people more towards using Sylt.

I looked over some of the comments, it looked mighty fine. :)

The bug should be added as a unit test

Quaqqer:
> It's needed if we want to distinguish between nested matches. > Consider this piece of code:
> ```sylt
> match 1
> with 1 -> match 2
>           with 2 -> amazing
> with 3 -> where_does_this_go
> ```

Parenthesis can be used for these cases, but sure, it may be unintuitive.

```sylt
match 1
  with 1 -> (match 2
    with 2 -> amazing)
  with 3 -> goes_here
```

FredTheDino:
https://github.com/sylt-lang/sylt-lang/pull/705#issuecomment-1425057712
The bug is not really a bug, more an oversight. Right now the definition order does not match that of Lua, the definition order in Sylt lofts declarations. One way to solve this is to simply not loft declarations. An easy fix but maybe not the behavior I would most want.

Regarding the parenthesis situation:
Yes, the ambiguity can be solved by adding parenthesis around your expression. So when I made this decision I had a few things in mind that made me want to have an `end` keyword.
 - I never ever want to care about white space - solutions with indentation are not an option
 - How much harm can this syntax change cause?

Consider this code yet again:
```
match a
with 1 -> match b
          with 2 -> amazing
with 3 -> where_does_this_go
```
One might write this code and think Sylt is indentation sensative - something I would not be suprised if people assumed. This causes people to think this program does something it doesn't do - namely when `a == 3` return `where_does_this_go`. This entire miscommunication can be solved by adding explicit `end`.

```
match a
with 1 -> match b
          with 2 -> amazing
          end
with 3 -> where_does_this_go
end
```
Now everyone will agree what this code does. This problem can be mitigated with totality checking on the match branches, but that is currently not implemented.

Parenthesis can be used to clear a lot of ambiguities, Lisp has already showed that, but it isn't always the best way to clear all ambiguities. I personally prefer to free up the parenthesis to be added here later. I also want people to always group their match-blocks to avoid confusion in the future.

The `def`s in Sylt functions like you describe:
```
def a ::= 1 + 1
def b ::= 1 + 1
```
So why am I okay with doing it like this but treating `match` differently?
Here the keywords that start each definition/type etc. is specifically picked to be unique. This means that anytime we see a `def` we have to been outside all functions, all scopes, and all definitions. The syntax is insanely simple thanks to this decision. Error recovery from invalid bodies can easily recover on the next definition. Since these `def`s cannot be nested, we cannot have the same ambiguity problems that we have with the match expressions.

In conclusion: Not grouping the match blocks can cause ambiguities and cause people to read the code differently. I want the syntax of the language to be explicit and I want to make this language something that is unlikely to bite you.

